PR #22, opened 2023-03-27
Update to use the current version of elfutils

godlygeek:


pablogsal:
Tested these wheels on my Arch Linux machine and the debuginfod support works 👍

PR #24, opened 2023-03-28
Test our wheels in more Linux distributions

pablogsal:
o increase the distribution coverage of our wheels, ensure that the wheels we generate work correctly with the interpreter shipped by Fedora, Debian, Ubuntu and Arch Linux.


godlygeek (approved):
LGTM

Issue #3, opened 2023-02-16
elfutils unwinding may be busted on Linux aarch64 when unwinding muslc


pablogsal:
When I try to unwind in an `Alpine Linux` docker container in a `aarch64` host seems that `elfutils` goes brrrrrrrrrrrrrrrrrrrrrrrrrrrr and chokes with an instruction pointer and keeps looping on it, so unwinding takes as much as the heat death of the universe.

We should try to add some magic sparkles so that unwinding doesn't take forever.

godlygeek:
I've also sent a bug report to the musl developers for this. https://www.openwall.com/lists/musl/2023/03/31/2

PR #26, opened 2023-03-30
Hack around infinite loop while unwinding on musl

godlygeek:
The elfutils library fails to successfully unwind multithreaded applications using musl libc. It gets stuck at the thread entry point (presumably due to poor debugging information in musl itself) and continually unwinds to the same instruction pointer forever.

Work around this with an ugly hack: when our callback is passed the same instruction pointer 5 times in a row, and the instruction pointer before those 5 was different, check to see if the function we've unwound to is called __clone and lives in ld-musl.so, and if so, break out of the loop on our own.

This should have minimal performance overhead for typical programs, and will not affect correctness when unwinding programs using glibc. The biggest risk is that there are some other cases where elfutils would begin to loop while unwinding musl code. That is: this has next to no chance of false positives, but does run a risk of false negatives.

Closes: #3

### DIFF
--- a/src/pystack/_pystack/unwinder.cpp
+++ b/src/pystack/_pystack/unwinder.cpp
@@ -74,6 +74,47 @@ DIENameFromScope(Dwarf_Die* die)
     return name == nullptr ? std::nullopt : std::optional<std::string>(name);
 }
 
+static bool
+breakMuslCloneLoop(Dwfl_Frame* state, std::vector<Frame>& frames)
+{
+    // elfutils gets stuck when unwinding multithreaded code on musl libc,
+    // passing a frame for __clone to our callback over and over forever.
+    // Detect if we've gotten into this state and break out if so.
+
+    // clang-format off
+    if (frames.size() < 5
+        || frames[frames.size() - 2] != frames.back()
+        || frames[frames.size() - 3] != frames.back()
+        || frames[frames.size() - 4] != frames.back()
+        || frames[frames.size() - 5] != frames.back()
+        || (frames.size() > 5 && frames[frames.size() - 6] == frames.back()))
+    {
+        return false;  // Not at the start of a run of 5+ identical frames
+    }
+    // clang-format on
+
+    Dwfl* dwfl = dwfl_thread_dwfl(dwfl_frame_thread(state));
+    Dwarf_Addr pc = frames.back().pc;
+    Dwarf_Addr pc_adjusted = pc - (frames.back().isActivation ? 0 : 1);
+    Dwfl_Module* mod = dwfl_addrmodule(dwfl, pc_adjusted);
+    const char* mod_name =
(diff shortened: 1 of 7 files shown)


godlygeek:
@pablogsal On #3 you said:

> For the time being I propose adding an unwind limit to the unwinder so at least we return something and don't loop forever. Together with #12 this should not be too expensive and the only problem is a lot of false `clone()` calls at the top of the stack.

I don't love that solution for a few reasons:
- thousands of `__clone` calls on the musl stacks sure is ugly, and
- that affects the behavior for glibc as well as musl.

We'd wind up truncating the stacks that we report for interpreters linked against glibc for a condition that we've only seen affecting musl. And we have seen stacks with millions of calls, when something manages to trigger infinite recursion in native code. Unconditionally throwing away valid stack frames rubs me the wrong way.

The biggest risk I see with doing it this way is that it might be too strict, and we might miss some other, similar, infinite loop. But I have tested, and this does work around the one that our test suite triggers.

pablogsal:
> Work around this with an ugly hack

 I think calling this ugly is being very generous 😆 But I agree that it may be our best bet to support this forsaken land of a platform 😓

pablogsal:
✨ The scope creep fairy has visited your PR and now it has been magically transformed into "Add Alpine support" ✨ 

You may present your complaints to the scope creep fairy office, opened only the second full moon of the year :)

src/pystack/_pystack/unwinder.cpp
```diff
@@ -74,6 +74,48 @@ DIENameFromScope(Dwarf_Die* die)
     return name == nullptr ? std::nullopt : std::optional<std::string>(name);
 }
 
+static bool
+breakMuslCloneLoop(Dwfl_Frame* state, std::vector<Frame>& frames)
```
pablogsal, line 78: Let's add some debug logging of some kind when this happens so we can detect what's going on if we end having reports around this

godlygeek: Good idea, done.

godlygeek:
> You may present your complaints to the scope creep fairy office

No complaints, this PR was in service of getting things passing on Alpine anyway.

> I think calling this ugly is being very generous 😆 But I agree that it may be our best bet to support this forsaken land of a platform 😓

Based on our offline discussion, I've switched it to break out if we ever see the same PC five times in a row inside musl libc, rather than looking specifically for `__clone`. This reduces our risk of false negatives (if there are other functions in musl libc that elfutils gets stuck on besides `__clone`, we'll avoid infinite looping even without knowing those cases in advance), but gives some risk of false positives (we might accidentally break out of a truly recursive function in the libc implementation).

pablogsal (approved):

godlygeek:
Flipping to draft so we don't land this before discussing #32.

godlygeek:
We're going to move forward with #32 instead.

PR #28, opened 2023-03-30
Harden `Py_TPFLAGS_*_SUBCLASS` checks

godlygeek:
Any actual class should have at most one of these subclass type flags set. If we see multiple of them set, it's safest to not trust that any of them are correct, and instead fall back to our opaque "other" handling.


godlygeek:
This fixes a failure I saw in an `ANONYMOUS_MAPS` test, where the test failed because it misidentified an object as a valid interpreter state, because the dictionaries it would have referenced if it was an interpreter state passed the `isValidDictionaryObject` despite very much not being valid dictionary objects (reference count around 2**63, `tp_name` not valid UTF-8, multiple of these `Py_TPFLAGS_FOO_SUBCLASS` flags set when interpreting the memory being held as `tp_flags`, etc).

I wonder if we should also fail that check if the `tp_name` isn't valid identifier characters... Or maybe even check for exactly `"dict"`...

pablogsal (approved):

PR #31, opened 2023-04-01
Allow overriding library paths from core files

godlygeek:
This enables analyzing a core file even if the libraries that were used by the application are not installed at the time PyStack is run (perhaps because the libraries on the machine have been updated). By providing PyStack with a set of directories to search for libraries in and placing the versions of the libraries that the application was using in those directories, users can analyze the core file without needing to modify the currently installed libraries on the machine.

### DIFF
--- a/src/pystack/_pystack/elf_common.cpp
+++ b/src/pystack/_pystack/elf_common.cpp
@@ -53,30 +53,6 @@ parse_permissions(long flags)
     return perms;
 }
 
-#if defined(BLOOMBERG_PYSTACK)
-static char*
-locateLibraryCallback(char* path, void* arg)
-{
-    assert(arg != nullptr);
-    CoreFileAnalyzer* analyzer = reinterpret_cast<CoreFileAnalyzer*>(arg);
-    if (!analyzer->d_executable || !analyzer->d_lib_search_path) {
-        if (!fs::exists(path)) {
-            analyzer->d_missing_modules.emplace_back(path);
-        }
-        return path;
-    }
-
-    std::string located_path = analyzer->locateLibrary(path);
-    if (!fs::exists(located_path)) {
-        analyzer->d_missing_modules.emplace_back(located_path);
-    }
-
-    char* the_name = strdup(located_path.c_str());
-    free(path);
-    return the_name;
-}
-#endif
-
 CoreFileAnalyzer::CoreFileAnalyzer(
         std::string corefile,
         std::optional<std::string> executable,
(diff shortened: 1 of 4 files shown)


src/pystack/_pystack/elf_common.cpp
```diff
@@ -144,13 +118,128 @@ CoreFileAnalyzer::~CoreFileAnalyzer()
     close(d_fd);
 }
 
+void
+CoreFileAnalyzer::removeModuleIf(std::function<bool(Dwfl_Module*)> predicate)
```
pablogsal, line 122: This can be made `const`

```suggestion
CoreFileAnalyzer::removeModuleIf(std::function<bool(Dwfl_Module*)> predicate) const
```

you need to update the header as well.

src/pystack/_pystack/elf_common.cpp
```diff
@@ -144,13 +118,128 @@ CoreFileAnalyzer::~CoreFileAnalyzer()
...
+    struct CallbackArgs
+    {
+        Dwfl* dwfl;
+        Predicate& predicate;
+    } callback_args = {d_dwfl.get(), predicate};
+
+    // Remove all modules, except for any that the callback re-adds.
+    dwfl_report_begin(d_dwfl.get());
+
+    int rc = dwfl_report_end(
```
pablogsal, line 134: Nit (feel free to ignore)

```suggestion
    const int rc = dwfl_report_end(
```

src/pystack/_pystack/elf_common.cpp
```diff
@@ -144,13 +118,128 @@ CoreFileAnalyzer::~CoreFileAnalyzer()
...
+        std::string located_path;
+        bool searched;
+        if (!d_executable || !d_lib_search_path) {
+            located_path = path;
+            searched = false;
+        } else {
+            located_path = locateLibrary(path);
+            searched = true;
+        }
+        bool located_path_exists = fs::exists(located_path);
```
pablogsal, line 188: Nit:

```suggestion
        const bool located_path_exists = fs::exists(located_path);
```

src/pystack/_pystack/elf_common.cpp
```diff
@@ -144,13 +118,128 @@ CoreFileAnalyzer::~CoreFileAnalyzer()
...
+
+        if (!located_path_exists) {
+            LOG(DEBUG) << "Adding " << path << " as a missing module "
+                       << (searched ? "despite" : "without") << " a search";
+            d_missing_modules.emplace_back(located_path);
+        }
+
+        if (located_path_exists && located_path != path) {
+            std::string filename = std::filesystem::path(located_path).filename().string();
```
pablogsal, line 197: Nit:

```suggestion
            const std::string filename = std::filesystem::path(located_path).filename().string();
```

pablogsal:
I have applied my suggestions and rebased for you :)

PR #32, opened 2023-04-02
Alternative hack around infinite loop while unwinding on musl

godlygeek:
An alternative approach to fixing the infinite loop we see while unwinding multithreaded musl code.

Closes: #3


godlygeek:
Compared to #26, this has a different set of tradeoffs. It should be faster, and it has essentially no risk of false positives - it is detecting exactly the condition that causes the problem. _But_, it means that we need a per-platform magic number sourced from the platform DWARF ABIs. The x86-64 value comes from table 3.18 on https://refspecs.linuxbase.org/elf/x86_64-abi-0.21.pdf and the aarch64 one from https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst#arm-specific-dwarf-definitions (actually, I pulled them both from the elfutils source code in `backends/*_initreg.c`, but that documentation confirms them). It also means we need at least elfutils 0.188 (that's when `dwfl_frame_reg` was added), and the version currently shipped on Alpine won't cut it - it's 0.187.

This is probably cleaner than the other fix, and assuming I understand the problem correctly it's probably the "right" fix, at least until one of the upstream projects fixes this somehow, but it isn't without its own set of disadvantages.

godlygeek:
I discussed this with @pablogsal offline, and he agreed that it's probably a better approach than #26, but the fact that a from-source build using the `elfutils` that ships with the latest Alpine would fail was a major downside. Based on that conversation, I've added a few more commits to this PR that build and test musllinux wheels. 🪄

src/pystack/_pystack/unwinder.cpp
```diff
@@ -83,7 +83,29 @@ frameCallback(Dwfl_Frame* state, void* arg)
     if (!dwfl_frame_pc(state, &pc, &isActivation)) {
         return -1;
     }
-    frames->emplace_back(pc, isActivation);
+
+    std::optional<Dwarf_Word> stackPointer;
+    std::optional<unsigned int> stackPointerRegNo;
+#if defined(__x86_64__)
```
pablogsal, line 89: Let's add some comments here indicating where the register numbers come from

pyproject.toml
```diff
@@ -42,3 +41,28 @@ before-all = [
   "CFLAGS='-w' ./configure --enable-libdebuginfod=dummy --disable-debuginfod --prefix=/usr --libdir=/usr/lib64",
   "make install"
 ]
+
+[[tool.cibuildwheel.overrides]]
+select = "*-musllinux*"
```
pablogsal, line 46: Maybe we can also add some comments here explaining this whole ordeal

pablogsal (approved):

PR #36, opened 2023-04-06
Show addresses in hex in VirtualMap repr

godlygeek:
This makes it much easier to cross reference our debug output against the data reported by `readelf` and `/proc/*/maps`.

### DIFF
--- a/src/pystack/maps.py
+++ b/src/pystack/maps.py
@@ -69,6 +69,17 @@ class VirtualMap:
     def size(self) -> int:
         return self.end - self.start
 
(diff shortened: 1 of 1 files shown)


godlygeek:
Example output:
```
INFO(parse_maps_file_for_binary): Heap map found: VirtualMap(start=0x00000000021d8000, end=0x00000000022ad000, filesize=0xd5000, offset=0x0, device='00:00', flags='rw-p', inode=0, path='[heap]')
```

pablogsal (approved):

PR #40, opened 2023-04-11
Drop hard dependency on elfutils 0.188

godlygeek:
This version of elfutils isn't available in most distributions yet, which makes a hard dependency on it (for `dwfl_frame_reg`) annoying. Since we only need such a new version for trying to avoid the infinite loop when gathering stacks for interpreters using musl libc, cheat and avoid performing the check when PyStack itself is built using glibc.

This isn't perfectly correct, but it should work for the overwhelming majority of cases, and lets us continue to be built with the versions of elfutils in most common distributions.

Update the README to call out that Alpine users will likely need to build elfutils from source in order to build PyStack from source, at least for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ Check your package manager on how to install these dependencies (e.g.,
 `apt-get install libdw-dev libelf-dev` in Debian-based systems). Note that you may need to tell the
 compiler where to find the header and library files of the dependencies for the build to
 succeed. Check your distribution's documentation to determine the location of the header and
-library files or for more detailed information.
+library files or for more detailed information. When building on Alpine Linux (or any other
+distribution that doesn't use glibc) you'll need elfutils 0.188 or newer. You may need to build this
+from source if your distribution's package manager doesn't have it.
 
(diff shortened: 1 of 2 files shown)


godlygeek:
Another option occurs to me here: we could make the check
```cpp
#if _ELFUTILS_VERSION >= 188 or (defined(__linux__) && !defined(__GLIBC__))
```

Then, we'd skip this check only if we're being built against an old version of elfutils, _and_ we're being built with a libc other than glibc. That might be a better set of compromises than this: it would mean that a PyStack built against glibc would still be able to unwind a Python process running with musl libc, as long as elfutils 0.188 or newer was available at build time.

pablogsal:
> Then, we'd skip this check only if we're being built against an old version of elfutils, _and_ we're being built with glibc. That might be a better set of compromises than this: it would mean that a PyStack built against glibc would still be able to unwind a Python process running with musl libc, as long as elfutils 0.188 or newer was available at build time.

I have pushed a commit implementing this check instead as I am convinced this is a slightly better scenario.

PR #74, opened 2023-04-24
Only delete the '*.so' that we create

### LINKED ISSUE #56: `make clean` breaks virtualenvs stored inside the repo clone
If you create a virtualenv inside a clone of the repo, running `make clean` will break it by removing all of the `*.so` files inside of it. We should improve our `make clean` to only remove files that we know we've created.

jay-z007:
*Issue number of the reported bug or feature request: #56 *

**Describe your changes**
Delete the following files:
- `_pystack`
- `build/` dir

**Testing performed**
Tested the virtualenv works after `make clean`

**Additional context**
Closes #56

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,10 @@ gh-pages:  ## Publish documentation on BBGitHub Pages
 
 .PHONY: clean
 clean:  ## Clean any built/generated artifacts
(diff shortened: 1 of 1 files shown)


godlygeek (approved):
LGTM

PR #75, opened 2023-04-24
Remove `extras-requires`

gusmonod:
Having these dependencies both in `requirements-*.txt` as well as in `setup.py` allows for inconsistencies, such as the missing `pyinstaller` dependency that really should be in the test dependencies. After discussing this, it seems like it's harder to read the `setup.py` file by Docker and CI in general than the `requirements-*.txt`.

**Describe your changes**
This PR removes all `extras_require` sections from `setup.py`, such that the development dependencies are not part of the package that gets published to PyPI.

**Testing performed**
Reinstalling and running all of the tests and coverage locally with that new setup.

**Additional context**
PyCon US 2023 ftw.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ install-sdist: dist  ## Install from source distribution
 
 .PHONY: test-install
 test-install:  ## Install with test dependencies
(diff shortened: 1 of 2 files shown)


godlygeek (approved):
LGTM

PR #77, opened 2023-04-24
Streamline linting process

stefmolin:
closes #72
closes #46 

**Describe your changes**
- Add `check-manifest`, `sphinx-lint`, and `clang-format` to pre-commit setup (removing from `make`)
- Remove intermediary `make` commands
- Clean up `Makefile` and C files with linting
- Remove extra requirements absorbed into pre-commit
- Configure pre-push hook to run in CI

**Testing performed**
Reinstalled the hooks to allow for pre-commit and pre-push hooks. Running `pre-commit run --all-files` runs in the pre-commit stage so all but `check-manifest`. Running `pre-commit run --all-files --hook-stage pre-push` only runs `check-manifest`.

PR #78, opened 2023-04-25
Replace flake8 and move isort on Python files to ruff

flpm:
Issue number of the reported bug or feature request: #45

**Describe your changes**
- Removed pre-commit check for flake8
- Removed setup.cfg (flake8 config)
- Added a pre-commit check for `ruff` that cover flake8 checks and isort for Python files
- Prevented isort pre-commit check from running on Python files

**Testing performed**
- Verified that ruff is fixing import order in .py and .pyi files (tested by introducing wrong order)
- Verified that ruff is fixing flake8 rules (tested by introducing unused import)

**Additional context**
Some `isort` options are not supported by `ruff`.

Option `multi_line_output = 3` (break multiple imports in multiple lines using parenthesis) is not supported by ruff. The default behavior is the same as `multi_line_output = 7` (all imports in one line without breaks), which used along `force-single-line = true` causes the multiple import to be split into individual imports. For example:
```
from collections import Counter, defaultdict, deque
```
is reformatted to:
```
from collections import Counter
from collections import defaultdict
from collections import deque
```
Options `include_trailing_comma = true` and `force_grid_wrap = 0` are related to `multi_line_output = 3` and are not applicable in ruff.

Option `use_parentheses = true` is also not supported by ruff, but because imports are in individual lines and `line-length = 95`, removing it would be very unlikely to cause problems.

Some references:
[Use Parentheses in isort documnentation](https://pycqa.github.io/isort/docs/configuration/options.html#use-parentheses)
[Multi Line Output Modes in isort documentation](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html)
[Flake 8 rules](https://www.flake8rules.com/)
[Ruff rules](https://beta.ruff.rs/docs/rules/)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,16 @@ requires = [
 
 build-backend = 'setuptools.build_meta'
 
-[tool.isort]
+[tool.ruff]
+line-length = 95
+select = ["F", "E", "W", "I001"]
 
+[tool.ruff.isort]
+force-single-line = true
+known-first-party = ["pystack"]
+known-third-party=["rich", "elftools", "pytest"]
+
+[tool.isort]
 force_single_line = true
 multi_line_output = 3
 include_trailing_comma = true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,23 +26,24 @@ repos:
       - id: rst-inline-touching-normal
         stages: [pre-commit]
 
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.262"
+    hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
+        stages: [pre-commit]
+
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
       - id: black
         stages: [pre-commit]
 
-  - repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
-    hooks:
-      - id: flake8
-        stages: [pre-commit]
-        types: [file, python]
-
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
     hooks:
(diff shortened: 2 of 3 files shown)


godlygeek:
@flpm Thank you! Would you mind editing your commit messages to add the [DCO signoff](https://github.com/bloomberg/pystack/blob/main/CONTRIBUTING.md#contribution-licensing)?

From the state your PR is in now, you can run:
```shell
git commit --amend -s
git push -f
```
to add a DCO signoff added to your most recent commit and then force-push it back to the PR.

flpm:
After PR #77 is merged I will rebase and fix the code, I will make sure all the commits have the DCO signoff :+1:

godlygeek:
I've merged #77 so you should now be unblocked, @flpm!

pyproject.toml
```diff
@@ -8,8 +8,16 @@ requires = [
...
+
+[tool.ruff.isort]
+force-single-line = true
+known-first-party = ["pystack"]
+known-third-party=["rich", "elftools", "pytest"]
 
+[tool.isort]
 force_single_line = true
 multi_line_output = 3
 include_trailing_comma = true
```
gusmonod, line 23: Do we feel strongly about keeping the import style the same as it is? I personally don't mind either way, as long as we have a consistent and enforceable style in the long run. Maybe we could remove `isort` altogether and use `ruff`'s convention instead?

It looks like these three options were introduced by Pablo when the project was started and never changed since then.

flpm: That’s right. The Makefile (before the cleanup) had flake8 running on Python files and isort running on both Python and Cython files.

godlygeek (approved):
LGTM

godlygeek:
Thanks for seeing this through, @flpm!

PR #91, opened 2023-04-28
ci: Test all built wheels

godlygeek:
Until now we were running all tests with the same `python3` version, but testing different Python versions for each matrix run.

Instead, install PyStack for each Python version from the built wheel, and only run the tests for that version.

Then, add a different test step that tests running PyStack with the latest Python version, and attaching to cores or processes running under old Python versions for which we don't have wheels.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -57,13 +57,13 @@ jobs:
         with:
           path: dist/*.tar.gz
 
-  test_wheels:
+  test_attaching_to_old_interpreters:
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
+        python_version: ["2.7", "3.6"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/download-artifact@v3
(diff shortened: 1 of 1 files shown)


godlygeek:
This is reviewable, but I'm leaving it as draft in order to rebase it on #92.

godlygeek:
#92 has landed and I've rebased, so this can be reviewed.

godlygeek:
In order to land this, we'll need to remove "test_wheels (3.6)" from the required checks list, and add "test_attaching_to_old_interpreters (3.6)"

gusmonod:
> Then, add a different test step that tests running PyStack with the latest Python version, and attaching to cores or processes running under old Python versions for which we don't have wheels.

Does this means that we don't test attaching to processes running under an older Python version for which we DO have wheels?

In other words, assuming 3.11 is latest, does this step test run `python3.11 -m pystack` on a list of processes each running with a Python version from 3.6 until 3.10 or only on a single process running under the Python 3.6 version?

godlygeek:
> Does this means that we don't want to test attaching to processes running under an older Python version for which we DO have wheels?

> In other words, should we add 3.7 until 3.10 to [this line](https://github.com/bloomberg/pystack/pull/91/files#diff-52d0610b43ce35e44510ae2d15d70668334378b01c1bdc774159cd3a33b71728R66)?

I went with "no" just to keep CI time down. We could, but it probably doesn't give us any interesting data beyond what we're getting from the other tests we're running. There's nothing particularly special about attaching to a different version of the interpreter than PyStack is running with; we can assume that the exact same code gets exercised regardless of whether you attach to a 3.7 process using a PyStack running under 3.7, or you attach to a 3.7 process using a PyStack running under 3.11.

The two things we most need to confirm are:
- We can successfully attach to every version we can run (2.7 and 3.6 through 3.11)
- We can run PyStack itself using every interpreter it supports (3.7 through 3.11)

gusmonod (approved):
LGTM

pablogsal:
I'm marking test_wheels 3.6 as not required so we can land this

PR #108, opened 2023-06-26
Support 3.12

godlygeek:


godlygeek:
This should be ready for a review now!

godlygeek:
There was a CI failure that I clicked rerun for. I'm pretty confident the flakiness is unrelated to these changes. Something is flaky somewhere, though.

src/pystack/_pystack/pyframe.cpp
```diff
@@ -44,14 +44,33 @@ FrameObject::FrameObject(
     d_prev = nullptr;
     d_next = nullptr;
 
+    d_is_shim = false;
```
pablogsal, line 47: The constructor is getting a bit complicated. What do you think if we start making small private functions to absorb the version-specific stuff something like:

```
uintptr_t FrameObject::getLastInstruction(const std::shared_ptr<const AbstractProcessManager>& manager,
                                          PyFrameObject& frame) {
    if (manager->majorVersion() > 3 || (manager->majorVersion() == 3 && manager->minorVersion() >= 11)) {
        return manager->getField(frame, &py_frame_v::o_prev_instr);
    } else {
        return manager->getField(frame, &py_frame_v::o_lasti);
    }
}

void FrameObject::checkForShim(const std::shared_ptr<const AbstractProcessManager>& manager,
                               PyFrameObject& frame) {
    d_is_shim = false;
    if (manager->majorVersion() > 3 || (manager->majorVersion() == 3 && manager->minorVersion() >= 12)) {
        constexpr int FRAME_OWNED_BY_CSTACK = 3;
        d_is_shim = manager->getField(frame, &py_frame_v::o_owner) == FRAME_OWNED_BY_CSTACK;
    }
}

void FrameObject::checkForEntry(const std::shared_ptr<const AbstractProcessManager>& manager) {
    if (manager->majorVersion() > 3 || (manager->majorVersion() == 3 && manager->minorVersion() >= 11)) {
        d_is_entry = false;  // Entry frames are detected below
    } else {
        d_is_entry = true;   // All frames are entry frames
    }
}
```

not sure if is worth it or is the time to do it though

src/pystack/_pystack/pyframe.cpp
```diff
@@ -44,14 +44,33 @@ FrameObject::FrameObject(
...
+        d_is_entry = true;  // All frames are entry frames
+    }
+
     if (d_prev_addr && d_frame_no < FRAME_LIMIT) {
         LOG(DEBUG) << std::hex << std::showbase << "Attempting to obtain new frame # " << d_frame_no + 1
                    << " from address " << d_prev_addr;
         d_prev = std::make_unique<FrameObject>(manager, d_prev_addr, frame_no + 1);
-        d_prev->d_next = this;
+        // Skip artificial shim frames inserted before entry frames in 3.12+
+        while (d_prev && d_prev->d_is_shim) {
```
pablogsal, line 64: For clarity with the other conditional, do we want to surround this with

```
if (manager->majorVersion() == 3 && manager->minorVersion() >= 12) {
```
?

src/pystack/_pystack/process.cpp
```diff
@@ -360,7 +359,14 @@ AbstractProcessManager::getStringFromAddress(remote_addr_t addr) const
         len = unicode._base._base.length;
         buffer.resize(len);
 
-        data_addr = ((remote_addr_t)((char*)addr + sizeof(PyASCIIObject)));
+        size_t offset;
+        if (d_major > 3 || (d_major == 3 && d_minor >= 12)) {
+            offset = sizeof(Python3_12::PyASCIIObject);
+        } else {
+            offset = sizeof(Python3::PyASCIIObject);
+        }
```
pablogsal, line 367: Maybe we can create some kind of structure that holds these types when a version is selected so we don't need to branch out in the code like this. What do you think?

pablogsal (approved):
Apart from the comments, this looks fantastic.

Great work, @godlygeek! ✨

src/pystack/_pystack/pythread.cpp
```diff
@@ -263,7 +260,27 @@ PyThread::calculateGilStatus(const std::shared_ptr<const AbstractProcessManager>
...
         LOG(DEBUG) << "_PyRuntime symbol detected. Searching for GIL status within _PyRuntime structure";
 
-        if (supportsExactGilChecking(manager->majorVersion(), manager->minorVersion())) {
+        if (manager->majorVersion() > 3
+            || (manager->majorVersion() == 3 && manager->minorVersion() >= 12))
+        {
+            // Fast, exact method supporting per-interpreter GILs:
+            // The thread state points to an interpreter state, which contains
+            // a ceval state, which points to a GIL runtime state.
+            // If `locked` is set and `last_holder` is us, we hold the GIL.
```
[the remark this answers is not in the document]
godlygeek: We is us!

Yeah, I'm referring to the thread represented by this `PyThread` instance. I can be more explicit.

src/pystack/_pystack/pyframe.cpp
```diff
@@ -44,14 +44,33 @@ FrameObject::FrameObject(
     d_prev = nullptr;
     d_next = nullptr;
 
+    d_is_shim = false;
+    if (manager->majorVersion() > 3 || (manager->majorVersion() == 3 && manager->minorVersion() >= 12)) {
+        constexpr int FRAME_OWNED_BY_CSTACK = 3;
```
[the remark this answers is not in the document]
godlygeek: Let's just leave it here for now. I think it's unlikely that we'll ever need to refer to this in more than one place, and it's tough to see how we should represent it our CPython headers until we see what things might look like in 3.13.

src/pystack/_pystack/pytypes.cpp
```diff
@@ -162,7 +162,18 @@ LongObject::LongObject(
 
     _PyLongObject longobj;
     manager->copyObjectFromProcess(addr, &longobj);
-    ssize_t size = longobj.ob_base.ob_size;
+    ssize_t size;
+    bool negative;
+
+    if (manager->majorVersion() > 3 || (manager->majorVersion() == 3 && manager->minorVersion() >= 12)) {
```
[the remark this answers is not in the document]
godlygeek: I badly wanted to improve this while working on this PR, but didn't want to make this already big PR any bigger. That said, now that you've reviewed this, I don't mind adding a commit on top of what you've already approved that improves this...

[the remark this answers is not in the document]
pablogsal: Let's do it 👍

src/pystack/_pystack/pytypes.cpp
```diff
@@ -162,7 +162,18 @@ LongObject::LongObject(
...
+        uintptr_t lv_tag = *reinterpret_cast<uintptr_t*>(&longobj.ob_base.ob_size);
+        negative = (lv_tag & 3) == 2;
+        size = lv_tag >> 3;
```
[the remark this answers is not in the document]
godlygeek: We're making an assumption here, but not about the behavior of this narrowing conversion. The `reinterpret_cast` above is assuming that `uintptr_t` and `ssize_t` are the same number of bits. As long as they are, that `reinterpret_cast` gives a `uintptr_t` value that's outside the range of an `ssize_t` if and only if the high bit is set, and then we right shift by 3 bits and the high bit can no longer be set, so the resulting value that we assign in the narrowing conversion must definitely be in range of an `ssize_t`. The behavior of narrowing conversions is well specified when the value being converted is in range for the type being converted to.

src/pystack/_pystack/pyframe.cpp
godlygeek: I'd rather not add an unnecessary conditional, honestly. `d_is_shim` will never be true unless the version is 3.12 or newer, so for older versions the body of this loop will never fire because the loop's condition will never be met. That seems clear enough to me.

src/pystack/_pystack/process.cpp
godlygeek: I'm not happy with this, but I couldn't spot a nicer way to write it that didn't seem to be adding a _lot_ of complexity. I started by defining the full structure of a `PyUnicodeObject` for each version, but I decided it was overkill and threw it away when I realized that it adds a fair amount of code, and that there's nothing we actually care about beyond the size of the `PyASCIIObject`, at least as things stand today. If you can see a nicer way to do this one, feel free to add a commit.

pablogsal: I tried something but is a lot of code and quite complex 😞

Let's leave it like this and hopefully we can find some better way to do this.

pablogsal: An idea for the future is having sone struct of sizes as well as one of offsets.

src/pystack/_pystack/pyframe.cpp
godlygeek: I spent a while trying to figure out a better way to factor the `FrameObject` constructor, and once I had it factored differently, it made sense to add this check. So, 👍 after all.

src/pystack/_pystack/process.cpp
godlygeek: > I tried something but is a lot of code and quite complex 😞

Yeah, I'll be honest: what you see in this PR was my 3rd attempt at this, and it's the least ugly one 😄. And the second least ugly had a magic `sizeof(wchar_t*)` in it, heh. I'm sure this could be improved, but I don't think it's easy, and I think this isn't too bad. I think my vote is to leave this alone until we have the next new `PyUnicodeObject` layout, and decide at that point how best to factor out the commonalities across them.

src/pystack/_pystack/pyframe.cpp
godlygeek: OK. I've added a commit that restructures this constructor to be implemented in terms of static helper functions that handle each of the version-specific things (getting the code object, checking if it's a shim frame, getting the previous frame, getting the `is_entry` flag). I think it's quite a lot better now. Let me know if you like it.

godlygeek:
I've rebased this, and I think I've responded to all your comments. Take another look when you can!

pablogsal (approved):

PR #110, opened 2023-06-30
Add a landing page for the documentation

pablogsal:
To make thinks shiny and cool, add a landing page instead of using the
Sphinx index page.


godlygeek:
Mm. We changed the GitHub Actions to not run the tests if nothing in src/ changed, which seemed clever at the time and like a way to save time - but of course, no our required status checks aren't reported 😓

godlygeek:
I've iterated on this a bit. I force pushed my changes, and the version with my changes can be previewed at https://godlygeek.github.io/pystack/

gusmonod:
About `docs/_static/images/native.png`, doesn't PyStack normally have different colors for C vs Python?

godlygeek:
> About `docs/_static/images/native.png`, doesn't PyStack normally have different colors for C vs Python?

Hm, yes...

> I'm also not a big fan of the red squiggles under some of the words

Also yes...

@pablogsal Would you like to take another swing at these? Or I can do it, but then we'll wind up with PuTTY instead of - iTerm2, I guess? We'll have Windows title bars instead of Mac ones, at least (or crop the title bars out 🤷)

pablogsal:
> > About `docs/_static/images/native.png`, doesn't PyStack normally have different colors for C vs Python?
> 
> 
> 
> Hm, yes...
> 
> 
> 
> > I'm also not a big fan of the red squiggles under some of the words
> 
> 
> 
> Also yes...
> 
> 
> 
> @pablogsal Would you like to take another swing at these? Or I can do it, but then we'll wind up with PuTTY instead of - iTerm2, I guess? We'll have Windows title bars instead of Mac ones, at least (or crop the title bars out 🤷)

Is not iterm, is https://carbon.sh. I can give it a go this night but if you want to replace them, just generate images with transparent backgrounds there

docs/_templates/index.html
```diff
@@ -0,0 +1,302 @@
...
+                </div><!-- /.col-lg-4 -->
+                <div class="col-lg-4">
+                    <img class="rounded-circle rounded-header-image"
+                        src="{{ pathto('_static/images/python.png', 1) }}"
+                        alt="Python logo" width="140" height="140">
+                    <h2>Ready for anything</h2>
+                    <p>
+                        PyStack thrives in extreme conditions, always ensuring
+                        accurate insights. It can handle deep call stacks like
+                        a breeze. It can get you an comprehensive report even if
```
gmc444-b, line 129:
```suggestion
                        a breeze. It can get you a comprehensive report even if
```

Alternatively, it could read "It can provide a comprehensive report even if..."

docs/_templates/index.html
```diff
@@ -0,0 +1,302 @@
...
+                        your Python interpreter is heavily optimized and shipped
+                        without debug information. It even copes with core files
+                        caused by memory corruption that may have trashed some
+                        of the interpreter's data structures.
```
gmc444-b, line 133: The last sentence leans towards alliteration which can degrade readability, and the punch line could be moved to the start, ie:
```suggestion
  Even if memory corruption has trashed some of the interpreter's internal data 
  structures, pystack can handle it!
```

docs/_templates/index.html
```diff
@@ -0,0 +1,302 @@
...
+                </div>
+            </div>
+
+            <hr class="featurette-divider">
+
+            <div class="row featurette">
+                <div class="col-md-7 order-md-2">
+                    <h2 class="featurette-heading">
+                        No more hidden code!
+                        <span class="text-muted">See into native extensions.</span>
```
gmc444-b, line 194: Perhaps a better synonym for "See", like "Peer into.." or "Dive into"

gmc444-b (commented):
A few suggestions.

Should one of the blurb mention that you'll also get stack traces for threads not owned by the interpreter? 

There's also this article that may be worthy of inclusion in the media section.

https://www.prweb.com/releases/bloomberg_wins_superuser_award_in_recognition_of_outstanding_use_support_of_open_infrastructure/prweb19385294.htm

docs/_templates/index.html
godlygeek: We have "Deep Dive into Program State." below... Let's go with:
```suggestion
                        <span class="text-muted">Peer inside native extensions.</span>
```

docs/_templates/index.html
godlygeek: I prefer "get you a" over "provide a" just to be a bit less formal (my writing tends to be overly formal, and part of my editing pass is always to try to sound at least a bit more folksy, haha)

docs/_templates/index.html
godlygeek: Hah, it would have been hard to be more alliterative if I had tried. Good catch!

pablogsal:
> There's also this article that may be worthy of inclusion in the media section.
> 
> [prweb.com/releases/bloomberg_wins_superuser_award_in_recognition_of_outstanding_use_support_of_open_infrastructure/prweb19385294.htm](https://www.prweb.com/releases/bloomberg_wins_superuser_award_in_recognition_of_outstanding_use_support_of_open_infrastructure/prweb19385294.htm)

Hummm, but this article does not speak itself about pystack directly no? Is more about BB in general more than focusing on the tool

lkollar (approved):

PR #113, opened 2023-07-10
Prepare for 1.1.0 release

pablogsal:
See changelog for more details.


NEWS.rst
```diff
@@ -8,6 +8,17 @@ Changelog
...
 
+pystack 1.1.0 (2023-07-10)
+--------------------------
+
+Bug Fixes
+~~~~~~~~~
+
+- Allow building with older elfutils than 0.188 when building with glibc (for musl libc we still need newer versions) (#40).
+- Improve error reporting when attaching to a process is forbidden (#98).
+- Fix a crash that could occur under some unusual conditions if elfutils could not unwind the stack (#101).
```
godlygeek, line 19: My fault, this should be:
```suggestion
- Allow building with older elfutils than 0.188 when building with glibc (for musl libc we still need newer versions). (#40)
- Improve error reporting when attaching to a process is forbidden. (#98)
- Fix a crash that could occur under some unusual conditions if elfutils could not unwind the stack. (#101)
```

godlygeek (approved):
LGTM

PR #117, opened 2023-07-11
Call `setup` from setuptools instead of distutils

godlygeek:
Also, ensure that `setuptools` is imported before `Cython`, per the Cython docs.


godlygeek:
This is a workaround for https://github.com/pypa/setuptools/issues/3984

pablogsal (approved):

PR #118, opened 2023-07-13
Fix error message for unsupported platforms

pablogsal:
closes: #122
closes: #121

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,6 @@ setuptools.setup(
         "Intended Audience :: Developers",
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: POSIX :: Linux",
-        "Operating System :: MacOS",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -1,10 +1,16 @@
...
 import sys
+from sys import platform
 
 import setuptools
 from Cython.Build import cythonize
 
+IS_LINUX = "linux" in platform
+
+if not IS_LINUX:
+    raise RuntimeError(f"memray does not support this platform ({platform})")
```
godlygeek, line 12: If the `setup.py` fails with a `RuntimeError`, won't `pip install pystack` just fall back to installing an older version that didn't have this check?

pablogsal: Hummmm, IIRC not anymore, but I am not sure. @pradyunsg do you know if this has changed or I am making this up?

pradyunsg: Not since https://pip.pypa.io/en/stable/news/#v22-0.

> Stop backtracking on build failures, by instead surfacing them to the user and aborting immediately. This behaviour provides more immediate feedback when a package cannot be built due to missing build dependencies or platform incompatibility. ([#10655](https://github.com/pypa/pip/issues/10655))

pablogsal: Thanks @pradyunsg!

Issue #54, opened 2023-04-23
Utilize `pkgconfig` to find elfutils


godlygeek:
The https://pypi.org/project/pkgconfig/ library provides a way to determine the right compiler and linker flags to use for building against a particular C library. We're not currently leveraging this, and are instead assuming that the elfutils headers and libraries are in the system's default search paths. We should update our `setup.py` to try to use `pkgconfig` to locate the right flags for building and linking against `libelf` and `libdw`, but fall back to our current behavior if the `pkgconfig` query fails.

jay-z007:
I'll take this

apurvakhatri:
pkgconfig raises EnvironmentError, If pkg-config is not on the path. (_As per pkgconfig Documentation_). How to include system package (pkg-config) in python project? If yes, what is the best practice? If no, should we update the docs for the user to add pkg-config to path before building? 
Also, I tested building editable by adding flags to PYSTACK_EXTENSION.extra_link_args & PYSTACK_EXTENSION.extra_compile_args. However, I am unable to understand the fallback behavior. How to go around implementing the fallback? Any help would be greatly appreciated. Thank you!

godlygeek:
> pkgconfig raises EnvironmentError, If pkg-config is not on the path. (_As per pkgconfig Documentation_). How to include system package (pkg-config) in python project? If yes, what is the best practice? If no, should we update the docs for the user to add pkg-config to path before building?

There's no way for a Python package to declare a dependency on a system package. We can mention that users should install `pkg-config` in the [Building from Source](https://github.com/bloomberg/pystack/blob/main/README.md#building-from-source) section of the README, though, like we do for `libdw` and `libelf` themselves.

> I am unable to understand the fallback behavior. How to go around implementing the fallback?

Just like you're doing - catching and ignoring the two possible exceptions that the `pkgconfig` module can raise (an `EnvironmentError` for `pkg-config` itself not being installed, and `pkgconfig.PackageNotFoundError` for it being installed but not being able to find or parse the `.pc` files). If either of those conditions happens, we just fall back to our current hardcoded `libraries=`.

> If you feel the linked PR is correct then please let me know. Any help would be greatly appreciated.

[...]

PR #120, opened 2023-07-16
Add pkgconfig support in the setup.py

apurvakhatri:
Issue number of the reported bug or feature request: #54 

**Describe your changes**
I have added support for pkgconfig in setup.py. Python's pkgconfig depends on system package pkg-config. The instructions on how to install pkg-config has also been added to README.md. The changes uses pkgconfig to look for clfags & ldflags to use for libelf and libdw. If pkgconfig is unable to determine the flags then the default pre-configured behavior is used. 

**Testing performed**
Testing has been performed on Ubuntu 22.04.2 LTS

**Additional context**
-

Closes: https://github.com/bloomberg/pystack/issues/54

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ package.
 Check your package manager on how to install these dependencies (e.g.,
 `apt-get install libdw-dev libelf-dev` in Debian-based systems). Note that you may need to tell the
 compiler where to find the header and library files of the dependencies for the build to succeed.
-This is handled at Pystack level provided the users install pkg-config (e.g.,
-`apt-get install pkg-config` in Debian-based systems).
+If `pkg-config` is available (e.g. `apt-get install pkg-config` on Debian-based systems), it will
+automatically be used to locate the libraries and configure the correct build flags.
 Check your distribution's documentation to determine the location of the header and library files
 or for more detailed information. When building on Alpine Linux (or any other distribution that
 doesn't use glibc) you'll need elfutils 0.188 or newer. You may need to build this from source if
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -64,6 +64,8 @@ package.
 Check your package manager on how to install these dependencies (e.g.,
 `apt-get install libdw-dev libelf-dev` in Debian-based systems). Note that you may need to tell the
 compiler where to find the header and library files of the dependencies for the build to succeed.
+This is handled at Pystack level provided the users install pkg-config (e.g.,
+`apt-get install pkg-config` in Debian-based systems).
```
godlygeek, line 68: Let's rephrase this a little bit:
```suggestion
If `pkg-config` is available (e.g. `apt-get install pkg-config` on Debian-based systems), it will
automatically be used to locate the libraries and configure the correct build flags.
```

setup.py
```diff
@@ -64,7 +77,7 @@
         "src/pystack/_pystack/unwinder.cpp",
         "src/pystack/_pystack/version.cpp",
     ],
-    libraries=["elf", "dw"],
+    libraries=library_flags['libraries'],
```
godlygeek, line 80: Doing it this way won't include any library directories or include directories reported by `pkg-config`. Instead, we should remove the `libraries=` entirely, and add `**library_flags` as the last argument to the `setup` call.

godlygeek (changes requested):
I see some stylistic issues that should be fixed by running `make format`, and a few other small issues as well:

godlygeek:
Ah, the Coverage workflow failed because https://github.com/bloomberg/pystack/blob/8473602e15afeb9130f21336e620bcbd2360bd12/.github/workflows/coverage.yml#L50 needs to install `pkgconfig`.

pablogsal:
I have rebased and manually applied @godlygeek suggestions to move this forward. Thanks a lot for the contribution @apurvakhatri 🚀

pablogsal (approved):

PR #124, opened 2023-07-18
LRU cache optimization 

ivonastojanovic:
A cache is added to reduce the number of system calls that copy the remote process's memory. The cache is a LRU cache with a specified capacity.


src/pystack/_pystack/mem.cpp
```diff
@@ -155,36 +158,156 @@ MemoryMapInformation::setHeap(const VirtualMap& heap)
...
+
+    auto it = cache.find(key);
+
+    if (it != cache.end()) {
+        cache_list.erase(it->second.second);
+        cache.erase(it);
+    }
+
+    while (size + value_size > cache_capacity) {
+        auto last = --cache_list.end();
```
godlygeek, line 182: It would be a little bit cleaner to use:
```suggestion
        auto last = cache_list.rbegin();
```
That gives you a reverse iterator to the first element in reverse order, so the last element in the list.

src/pystack/_pystack/mem.cpp
```diff
@@ -155,36 +158,156 @@ MemoryMapInformation::setHeap(const VirtualMap& heap)
...
+        cache.erase(last->first);
+        size -= last->second;
+        cache_list.pop_back();
+    }
+
+    cache_list.push_front(list_node_t{key, size});
```
godlygeek, line 188: Shouldn't this be:
```suggestion
    cache_list.push_front(list_node_t{key, value_size});
```

godlygeek (changes requested):
This looks pretty good! I do seem some issues that will need to be cleaned up, and I have a few thoughts. Also, linting failed, so there's some stylistic issues. You should be able to fix those by [setting up `pre-commit`](https://github.com/bloomberg/pystack/blob/main/CONTRIBUTING.md#linting-your-code). Once you've got that all set up, `make format` should format the code to be compliant with our CI checks, and `git commit` will complain unless the code is properly formatted.

godlygeek:
I'm not sure if any of the bugs I noticed were the cause, but the coverage stage of CI failed with:
```
INFO     /home/runner/work/pystack/pystack/src/pystack/_pystack.cpython-310-x86_64-linux-gnu.so:test_gather_stacks.py:119 Fetching Python threads
double free or corruption (out)
Fatal Python error: Aborted
```

We'll have to see if the bugs I noticed account for that, or if that's an unrelated issue...

godlygeek:
Lastly, the "Commits Check" check is ensuring that you've provided a [DCO sign-off](https://github.com/bloomberg/pystack/blob/main/CONTRIBUTING.md#contribution-licensing) in your commit message. You could add that to your current commit by doing a `git commit --amend -s` and then force-pushing your changed commit up to GitHub with `git push -f`

src/pystack/_pystack/mem.cpp
```diff
@@ -155,36 +156,137 @@ MemoryMapInformation::setHeap(const VirtualMap& heap)
...
         }
-        throw std::system_error(errno, std::generic_category());
-    }
+
+        result += read;
+    } while ((size_t)read != local[0].iov_len);
 
     if ((size_t)result != len) {
         throw InvalidCopiedMemory();
     }
```
godlygeek: Unless I'm missing something, it doesn't seem like this can happen anymore - if we've broken out of the `while` loop, it's because `read == local[0].iov_len`, and `local[0].iov_len` was the amount we needed to read before `result` would equal `len`, so if `read == local[0].iov_len`, then `result == len`.

```suggestion
```

godlygeek (commented):
This all looks good to me at this point. I don't see any more bugs. I've got some nitpicks, but rather than force you to go through another round of reviews, I'm just going to push a fixup commit that improves those little things.

@ivonastojanovic Please check my work and make sure I didn't get anything wrong or make anything worse in my fixup commit.

@pablogsal Would you like to review this as well before we land it?

ivonastojanovic:
> This all looks good to me at this point. I don't see any more bugs. I've got some nitpicks, but rather than force you to go through another round of reviews, I'm just going to push a fixup commit that improves those little things.
> 
> @ivonastojanovic Please check my work and make sure I didn't get anything wrong or make anything worse in my fixup commit.
> 
> @pablogsal Would you like to review this as well before we land it?

ivonastojanovic:
@godlygeek Everything looks great!

src/pystack/_pystack/mem.h
```diff
@@ -109,6 +110,28 @@ class MemoryMapInformation
     std::optional<VirtualMap> d_heap;
 };
 
+class LRUCache
+{
+  private:
+    using list_node_t = std::pair<uintptr_t, size_t>;
```
pablogsal, line 116: Just a small nit: in general I find expressions like `it->second.second` a bit challenging to check for correctness and it always forces me to go to the definitions. Given that we would not be sacrificing performance or anything I recommend here to use a names struct for `list_node_t` so we can use the names directly instead of the position on the pair.

The same thing applies to `cache_value_t`

pablogsal (approved):
LGTM

PR #126, opened 2023-07-28
Mapping the core file

ivonastojanovic:
Changed the way of reading data from the core file. The core file is mapped directly into a virtual memory page.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/pystack/_pystack/mem.cpp
+++ b/src/pystack/_pystack/mem.cpp
@@ -344,23 +344,89 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
 {
     CoreFileExtractor extractor{d_analyzer};
     d_shared_libs = extractor.ModuleInformation();
+
+    const char* filename = d_analyzer->d_filename.c_str();
+    int fd = open(filename, O_RDONLY);
+
+    if (fd == -1) {
+        LOG(ERROR) << "Failed to open a file " << filename;
+        throw RemoteMemCopyError();
+    }
+
+    StatusCode ret = readCorefile(fd, filename);
+    int close_ret = close(fd);
+
+    if (close_ret == -1) {
+        LOG(ERROR) << "Failed to close a file " << filename;
+        throw RemoteMemCopyError();
+    }
+
+    if (ret == StatusCode::ERROR) {
+        throw RemoteMemCopyError();
+    }
+}
+
+CorefileRemoteMemoryManager::StatusCode
+CorefileRemoteMemoryManager::readCorefile(int fd, const char* filename) noexcept
+{
+    struct stat fileInfo = {0};
+
+    if (fstat(fd, &fileInfo) == -1) {
+        LOG(ERROR) << "Failed to get a file size for a file " << filename;
+        return StatusCode::ERROR;
+    }
+
+    if (fileInfo.st_size == 0) {
+        LOG(ERROR) << "File " << filename << " is empty";
+        return StatusCode::ERROR;
+    }
+
+    d_corefile_size = fileInfo.st_size;
+
+    void* map = mmap(0, d_corefile_size, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (map == MAP_FAILED) {
+        LOG(ERROR) << "Failed to mmap a file " << filename;
+        return StatusCode::ERROR;
+    }
+
+    d_corefile_data = std::unique_ptr<char, std::function<void(char*)>>(
+            reinterpret_cast<char*>(map),
+            [this](auto addr) {
+                if (munmap(addr, d_corefile_size) == -1) {
+                    LOG(ERROR) << "Failed to un-mmap a file " << d_analyzer->d_filename.c_str();
+                }
+            });
+
+    int madvise_result = madvise(d_corefile_data.get(), d_corefile_size, MADV_RANDOM);
+
+    if (madvise_result == -1) {
+        LOG(WARNING) << "Madvise for a file " << filename << " failed";
+    }
+
+    return StatusCode::SUCCESS;
 }
 
 ssize_t
 CorefileRemoteMemoryManager::copyMemoryFromProcess(remote_addr_t addr, size_t size, void* destination)
         const
 {
-    const std::string* filename = nullptr;
     off_t offset_in_file = 0;
 
-    StatusCode ret = getMemoryLocationFromCore(addr, &filename, &offset_in_file);
+    StatusCode ret = getMemoryLocationFromCore(addr, &offset_in_file);
 
-    if (ret == StatusCode::ERROR) {
-        // The memory may be in the data segment of some shared library
-        getMemoryLocationFromElf(addr, &filename, &offset_in_file);
+    if (ret == StatusCode::SUCCESS) {
+        if (static_cast<size_t>(offset_in_file) > d_corefile_size) {
+            throw InvalidRemoteAddress();
+        }
+        memcpy(destination, d_corefile_data.get() + offset_in_file, size);
+        return size;
     }
 
-    if (filename == nullptr) {
+    // The memory may be in the data segment of some shared library
+    const std::string* filename = nullptr;
+    ret = getMemoryLocationFromElf(addr, &filename, &offset_in_file);
+
+    if (ret == StatusCode::ERROR) {
         throw InvalidRemoteAddress();
     }
 
@@ -376,10 +442,7 @@ CorefileRemoteMemoryManager::copyMemoryFromProcess(remote_addr_t addr, size_t si
 }
 
 CorefileRemoteMemoryManager::StatusCode
-CorefileRemoteMemoryManager::getMemoryLocationFromCore(
-        remote_addr_t addr,
-        const std::string** filename,
-        off_t* offset_in_file) const
+CorefileRemoteMemoryManager::getMemoryLocationFromCore(remote_addr_t addr, off_t* offset_in_file) const
 {
     auto corefile_it = std::find_if(d_vmaps.cbegin(), d_vmaps.cend(), [&](auto& map) {
         return (map.Start() <= addr && addr <= map.End()) && (map.FileSize() != 0 && map.Offset() != 0);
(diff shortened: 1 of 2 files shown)


pablogsal:
Some local benchmarks:

# Before
```
root@fdfb53906b9d:/src# time pystack core core.630 --locals > /dev/null

real	0m0.374s
user	0m0.098s
sys	0m0.141s
```
# After
```
root@fdfb53906b9d:/src# time pystack core core.630 --locals > /dev/null

real	0m0.131s
user	0m0.076s
sys	0m0.043s
```

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
 {
     CoreFileExtractor extractor{d_analyzer};
     d_shared_libs = extractor.ModuleInformation();
+
+    const char* filename = d_analyzer->d_filename.c_str();
+    int fd = open(filename, O_RDONLY);
+
+    StatusCode ret = readCorefile(fd, filename);
+    close(fd);
```
pablogsal, line 352: Close can fail as well, we should log the problem

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
+
+CorefileRemoteMemoryManager::~CorefileRemoteMemoryManager()
+{
+    if (munmap(corefile_data, corefile_size) == -1) {
+        LOG(ERROR) << "Failed to un-mmap a file " << d_analyzer->d_filename.c_str();
+    }
+}
+
+CorefileRemoteMemoryManager::StatusCode
+CorefileRemoteMemoryManager::readCorefile(int fd, const char* filename)
```
pablogsal, line 367: Let's mark this `noexcept` to make it clear that the `close` call before will always be called because this function does not raise

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
+    // The memory may be in the data segment of some shared library
+    const std::string* filename = nullptr;
+    ret = getMemoryLocationFromElf(addr, &filename, &offset_in_file);
+
+    if (ret == StatusCode::ERROR) {
         throw InvalidRemoteAddress();
     }
 
     std::ifstream is(*filename, std::ifstream::binary);
+
```
pablogsal, line 426: Straneous new line

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
     if (is) {
         is.seekg(offset_in_file);
         is.read((char*)destination, size);
     } else {
         LOG(ERROR) << "Failed to read memory from file " << *filename;
         throw InvalidRemoteAddress();
     }
+
```
pablogsal, line 434: Straneous new line

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
-    const std::string* filename = nullptr;
     off_t offset_in_file = 0;
 
-    StatusCode ret = getMemoryLocationFromCore(addr, &filename, &offset_in_file);
+    StatusCode ret = getMemoryLocationFromCore(addr, &offset_in_file);
 
-    if (ret == StatusCode::ERROR) {
-        // The memory may be in the data segment of some shared library
-        getMemoryLocationFromElf(addr, &filename, &offset_in_file);
+    if (ret == StatusCode::SUCCESS) {
```
pablogsal, line 412: Let's add an assert here that offset_in_file is smaller than the mmaped chunk

src/pystack/_pystack/mem.h
```diff
@@ -227,11 +232,11 @@ class CorefileRemoteMemoryManager : public AbstractRemoteMemoryManager
     std::shared_ptr<CoreFileAnalyzer> d_analyzer;
     std::vector<VirtualMap> d_vmaps;
     std::vector<SimpleVirtualMap> d_shared_libs;
+    char* corefile_data;
```
pablogsal, line 235: @godlygeek I advised to start with the `CorefileRemoteMemoryManager` owning this pointer instead of encapsulating it into a RAII based thing, but this is something I wanted to check with you. We could move all the mmap logic to its own layer and handle the thing in a destructor so this class just holds a `unique_ptr` and that's it. Thoughts?

godlygeek: Hm. I think it's fine having the destructor do it, but it doesn't seem like it would be hard to have RAII without adding a new mmap layer, unless I'm missing something. Can't we just declare it as:
```cpp
    std::unique_ptr<char, std::function<void(char*)>> corefile_data;
```
and initialize it as:
```cpp
    corefile_data = std::unique_ptr<char, std::function<void(char*)>>(
            reinterpret_cast<char*>(mmap(0, corefile_size, PROT_READ, MAP_PRIVATE, fd, 0)),
            [this](auto addr) {
                if (munmap(addr, corefile_size) == -1) {
                    LOG(ERROR) << "Failed to un-mmap a file " << d_analyzer-d_filename.c_str();
                }
            });
```
Though we'd need to sprinkle `.get()` everywhere where we use the pointer, which is a bit annoying, so 🤷 I guess on balance I'm fine with just calling `munmap` in the destructor.

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
+    close(fd);
+
+    if (ret == StatusCode::ERROR) {
+        throw RemoteMemCopyError();
+    }
+}
+
+CorefileRemoteMemoryManager::~CorefileRemoteMemoryManager()
+{
+    if (munmap(corefile_data, corefile_size) == -1) {
```
pablogsal, line 361: If `mmap` in the constructor failed we need to avoid this call

src/pystack/_pystack/mem.h
```diff
@@ -227,11 +232,11 @@ class CorefileRemoteMemoryManager : public AbstractRemoteMemoryManager
...
+    size_t corefile_size;
```
godlygeek, line 236: The other member data have a `d_` prefix on their names. We should match that naming convention here:
```suggestion
    char* d_corefile_data;
    size_t d_corefile_size;
```

godlygeek:
https://github.com/bloomberg/pystack/commit/1da05550d22a17debeea7fba25bfe0585e68cc34 sets up RAII for this, and I think it's a small improvement over what you've got - @ivonastojanovic would you mind integrating that into your branch? Specifically:
- the data type for the mapping member variable becomes `std::unique_ptr<char, std::function<void(char*)>>`
- we provide a lambda as the deleter when assigning it
- that deleter calls munmap, and logs if it fails
- we check for a map error before assigning it (since we don't want to call munmap in that case)
- we flip the order of the two member variables (C++ member variables are destroyed in reverse order, and we need `d_corefile_data` to be destroyed before `d_corefile_size`, since the lambda that calls munmap needs to refer to `d_corefile_size`)

src/pystack/_pystack/mem.cpp
```diff
@@ -344,42 +344,99 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
+        LOG(ERROR) << "Failed to mmap a file " << filename;
+        return StatusCode::ERROR;
+    }
+
+    int madvise_result = madvise(corefile_data, corefile_size, MADV_RANDOM);
+
+    if (madvise_result == -1) {
+        LOG(ERROR) << "Madvise for a file " << filename << " failed";
+        return StatusCode::ERROR;
+    }
```
godlygeek, line 399: I don't think we need to treat this as a fatal error. `madvise` shouldn't fail, but it doesn't matter if it does. It's purely advisory and affects only performance, so if this call fails, we should probably log a warning but otherwise continue as though it had succeeded.

src/pystack/_pystack/mem.cpp
godlygeek: Should it be an assert, or should we `throw InvalidRemoteAddress()`? I'm not sure if that's an invariant or something that depends on the contents of the core file...

src/pystack/_pystack/mem.cpp
```diff
@@ -344,23 +344,89 @@ CorefileRemoteMemoryManager::CorefileRemoteMemoryManager(
...
     off_t offset_in_file = 0;
 
-    StatusCode ret = getMemoryLocationFromCore(addr, &filename, &offset_in_file);
+    StatusCode ret = getMemoryLocationFromCore(addr, &offset_in_file);
 
-    if (ret == StatusCode::ERROR) {
-        // The memory may be in the data segment of some shared library
-        getMemoryLocationFromElf(addr, &filename, &offset_in_file);
+    if (ret == StatusCode::SUCCESS) {
+        if (static_cast<size_t>(offset_in_file) > d_corefile_size) {
```
godlygeek, line 418: This should be checking that the entire range being copied lies within the mapped file, not just the start of the range. That is, I think this should be:
```suggestion
        if (size > d_corefile_size || static_cast<size_t>(offset_in_file) > d_corefile_size - size) {
```

src/pystack/_pystack/mem.h
```diff
@@ -227,11 +231,11 @@ class CorefileRemoteMemoryManager : public AbstractRemoteMemoryManager
...
     std::vector<VirtualMap> d_vmaps;
     std::vector<SimpleVirtualMap> d_shared_libs;
+    size_t d_corefile_size;
+    std::unique_ptr<char, std::function<void(char*)>> d_corefile_data;
 
-    StatusCode getMemoryLocationFromCore(
-            remote_addr_t addr,
-            const std::string** filename,
-            off_t* offset_in_file) const;
+    StatusCode readCorefile(int fd, const char* filename) noexcept;
```
godlygeek, line 237: Let's rename this to `mapCoreFile` instead of `readCoreFile` to make it clearer what it's doing.
```suggestion
    StatusCode mapCorefile(int fd, const char* filename) noexcept;
```

godlygeek (approved):
I've noticed just two small issues here on re-review. I'm going to force-push a commit changing both of those small things, to avoid another round of reviews. Otherwise, this LGTM!

I'm not going to land this yet, in case @pablogsal wants to take another look.

pablogsal (approved):

PR #129, opened 2023-08-02
ci: Automatically publish docs for new releases

pablogsal:
On each new release, automatically build our docs and publish them to GitHub Pages. The docs are also built by our linting workflow, but we don't attempt to reuse that build here, because the linting build runs `towncrier` with a phony version number in order to exercise the news fragments in the news/ directory, and we don't want that fake version in the changelog we publish.


godlygeek (approved):
I uncommented the `if: github.event_name == 'release' && github.event.action == 'published'` so that this will only run on new releases, and force-pushed. Otherwise, LGTM.

PR #131, opened 2023-08-09
ci: Update already installed Arch Linux packages

godlygeek:
We seem to be installing a version of `libstdc++` that's incompatible with the version of glibc that the image shipped with.

The Arch Linux wiki says to avoid using `pacman -Sy` and to instead use `pacman -Syu`. Issues like this are probably why.


godlygeek:
CI passes, so - it works 😄

gusmonod (approved):

PR #136, opened 2023-08-18
Add link in the README to the Arch User Repository

Rubo3:
This patch adds a mention in the README to the PKGBUILD on the Arch User Repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ PyStack has the following amazing features:
 
 At this time only Linux is supported.
 
+It can be found on the [Arch User Repository](https://aur.archlinux.org/packages/python-pystack).
+
 ## Building from source
 
 If you wish to build PyStack from source, you need the following binary dependencies in your


gusmonod:
The AUR is specific to Arch Linux, even though `pystack` supports other distros in addition to Arch Linux.

@godlygeek do we support publishing binaries to distributions' repositories, or do we only publish to PyPI and expect distribution owners to decide whether or not to add them there?

godlygeek:
> @godlygeek do we support publishing binaries to distributions' repositories, or do we only publish to PyPI and expect distribution owners to decide whether or not to add them there?

That'd be up to distro maintainers to do. I guess I'm not against updating the README to include mention of distributions that package PyStack, but I don't think it belongs in the "## What platforms are supported?" section of the README. I guess the right thing to do is probably creating a new Installation section, and explaining there that you can either install it via `pip install`, or install it from your distribution's repositories if your distribution packages PyStack.

gusmonod:
@Rubo3 would you be interested in doing what godlygeek described above? Adding the Installation section that is.

gusmonod:
Closing as this has had no activity recently. @Rubo3 feel free to comment on this again if you are willing to look into it after all.

Rubo3:
Oh yes sorry, I'll make it.

Rubo3:
Done. You can re-open this PR if you want.

PR #139, opened 2023-08-22
Add missing doc for --self flag

alicederyn:
The docs for `pystack remote --self` appear to have been missed off the README, though the `--self` flag is shown in the usage string. Copy/pasted from the output of `pystack remote --help`

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ options:
   --native-all   Include native (C) frames from threads not registered with the interpreter (implies --native)
   --locals       Show local variables for each frame in the stack trace
   --exhaustive   Use all possible methods to obtain the Python stack info (may be slow)
+  --self         Introspect the same process that invoke this program
 ```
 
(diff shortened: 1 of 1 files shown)


alicederyn:
I tried to open an issue but none of the templates made any sense for a docs issue.

godlygeek:
I think we should probably just remove this flag. It isn't exercised by our tests, and it appears to have bit rotted:
```sh
$ pystack remote --self
usage: pystack remote [-h] [-v] [--no-color] [--no-block] [--native] [--native-all] [--locals] [--exhaustive] [--self] pid
pystack remote: error: the following arguments are required: pid
```

I think we removed the code for handling this flag at some point, and missed removing the flag itself.

@alicederyn Would you be interested in updating the PR to remove all mentions of `--self` instead?

alicederyn:
Closing in favour of #141.

PR #152, opened 2023-11-16
Fix CI target for Fedora

pablogsal:
Seems that the latest fedora image ships Python 3.12 with not enough
debugging symbols to provide native information for core files.
Apparently the python3-debug package only provides the debug version of
the runtime and doesn't also provide split-debug information for the
main executable.

Luckily, Fedora runs a debuginfod server that we can leverage to obtain
the debugging symbols that we need.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -204,8 +204,9 @@ jobs:
             g++ \
             file \
             python3 \
-            python3-devel \
-            python3-debug
+            python3-devel
+          dnf debuginfo-install -y \
+            python3
       - name: Install Python dependencies
         run: |
           python3 -m pip install --upgrade pip
--- a/src/pystack/_pystack/pythread.cpp
+++ b/src/pystack/_pystack/pythread.cpp
@@ -96,9 +96,20 @@ findPthreadTidOffset(
         auto pthread_id_addr = manager->getField(current_thread, &py_thread_v::o_thread_id);
 
         // Attempt to locate a field in the pthread struct that's equal to the pid.
-        uintptr_t buffer[200];
-        manager->copyObjectFromProcess(pthread_id_addr, &buffer);
-        for (int i = 0; i < 200; i++) {
+        uintptr_t buffer[100];
+        size_t buffer_size = sizeof(buffer);
+        while (buffer_size > 0) {
+            try {
+                LOG(DEBUG) << "Trying to copy a buffer of " << buffer_size << " bytes to get pthread ID";
+                manager->copyMemoryFromProcess(pthread_id_addr, buffer_size, &buffer);
+                break;
+            } catch (const RemoteMemCopyError& ex) {
+                LOG(DEBUG) << "Failed to copy buffer to get pthread ID";
+                buffer_size /= 2;
+            }
+        }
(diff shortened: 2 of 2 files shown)


pablogsal:
@hroncok do you know if there is any other way to obtain debugging symbols for the main executable of CPython in Fedora other than using `debuginfod`?

hroncok:
Something like `dnf debuginfo-install python3-libs python3`

.github/workflows/build_wheels.yml
```diff
@@ -206,6 +208,9 @@ jobs:
             python3 \
             python3-devel \
             python3-debug
```
hroncok, line 210: Unless you run python3-debug, you don't need it.
```suggestion
```

.github/workflows/build_wheels.yml
```diff
@@ -206,6 +208,9 @@ jobs:
...
+          dnf debuginfo-install -y \
+            python3-libs \
+            python3
```
hroncok, line 213: Looking at the log apparently listing both is redundant:

```suggestion
            python3
```

src/pystack/_pystack/pythread.cpp
```diff
@@ -96,9 +96,20 @@ findPthreadTidOffset(
         auto pthread_id_addr = manager->getField(current_thread, &py_thread_v::o_thread_id);
 
         // Attempt to locate a field in the pthread struct that's equal to the pid.
-        uintptr_t buffer[200];
-        manager->copyObjectFromProcess(pthread_id_addr, &buffer);
-        for (int i = 0; i < 200; i++) {
+        uintptr_t buffer[100];
+        size_t buffer_size = sizeof(uintptr_t) * 100;
```
godlygeek, line 100: it's a nit, but I strongly prefer `size_t buffer_size = sizeof(buffer);` - it removes duplication (`100` twice, `uintptr_t` twice) and thus eliminates the chances of a mistake that a compiler won't catch if we adjust one of those.
```suggestion
        size_t buffer_size = sizeof(buffer);
```

src/pystack/_pystack/pythread.cpp
```diff
@@ -96,9 +96,20 @@ findPthreadTidOffset(
...
+        while (buffer_size >= 0) {
```
godlygeek, line 101: Surely this should be `> 0`? We don't want to keep looping if reading 1 byte failed.
```suggestion
        while (buffer_size > 0) {
```

src/pystack/_pystack/pythread.cpp
```diff
@@ -96,9 +96,20 @@ findPthreadTidOffset(
...
+            try {
+                LOG(DEBUG) << "Trying to copy a buffer of " << buffer_size << " bytes to get pthread ID";
+                manager->copyMemoryFromProcess(pthread_id_addr, buffer_size, &buffer);
+                break;
+            } catch (const RemoteMemCopyError& ex) {
+                LOG(DEBUG) << "Failed to copy buffer to get pthread ID: ";
```
godlygeek, line 107:
```suggestion
                LOG(DEBUG) << "Failed to copy buffer to get pthread ID";
```

godlygeek (approved):
LGTM

PR #161, opened 2023-12-15
Update `upload-artifact` and `download-artifact`

gusmonod:
Replace #160 which was failing because of breaking changes in v4.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -24,8 +24,9 @@ jobs:
       - name: Build sdist
         run: pipx run build --sdist
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
+          name: sdist
           path: dist/*.tar.gz
 
   choose_architectures:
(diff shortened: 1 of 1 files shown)


gusmonod:
To the reviewer(s):

1. I haven't tested `aarch64` at all because I can't think of a reason why it wouldn't work. Let me know if you want a test similar to the one I'm describing below.
2. I've changed the `upload_pypi` to run unconditionally so that you can check that the `dist/` directory contains the right things. Once you're convinced it works as expected, we can apply the following patch to make it run only under the appropriate circumstances:

```patch
diff --git a/.github/workflows/build_wheels.yml b/.github/workflows/build_wheels.yml
index cf135aa..23be339 100644
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -306,7 +306,7 @@ jobs:
   upload_pypi:
     needs: [test_wheels]
     runs-on: ubuntu-latest
-#   if: github.event_name == 'release' && github.event.action == 'published'
+    if: github.event_name == 'release' && github.event.action == 'published'
     steps:
       - uses: actions/download-artifact@v4
         with:
@@ -316,8 +316,7 @@ jobs:
           mv dist/sdist/*.tar.gz dist/
           mv dist/wheels-*/*.whl dist/
           rmdir dist/{sdist,wheels-*}
-      - run: ls -R dist
-#     - uses: pypa/gh-action-pypi-publish@release/v1
-#       with:
-#         skip_existing: true
-#         password: ${{ secrets.PYPI_PASSWORD }}
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          skip_existing: true
+          password: ${{ secrets.PYPI_PASSWORD }}
```

gusmonod:
Closing in favor of #163

PR #163, opened 2024-01-02
Update `upload-artifact` and `download-artifact`: one artifact per arch + platform tag series

gusmonod:
Another approach for #161 where we use one artifact per architecture + platform tag series instead of one per wheel.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -29,26 +29,27 @@ jobs:
           name: sdist
           path: dist/*.tar.gz
 
-  choose_architectures:
-    name: Decide which architectures to build wheels for
+  choose_wheel_types:
+    name: Decide which wheel types to build
     runs-on: ubuntu-latest
     steps:
-      - id: x86_64
-        run: echo "cibw_arch=x86_64" >> $GITHUB_OUTPUT
-      - id: aarch64
-        run: echo "cibw_arch=aarch64" >> $GITHUB_OUTPUT
+      - id: manylinux_x86_64
+        run: echo "wheel_types=manylinux_x86_64" >> $GITHUB_OUTPUT
+      - id: musllinux_x86_64
+        run: echo "wheel_types=musllinux_x86_64" >> $GITHUB_OUTPUT
+      - id: manylinux_aarch64
+        run: echo "wheel_types=manylinux_aarch64" >> $GITHUB_OUTPUT
     outputs:
-      cibw_arches: ${{ toJSON(steps.*.outputs.cibw_arch) }}
+      wheel_types: ${{ toJSON(steps.*.outputs.wheel_types) }}
 
   build_wheels:
-    needs: [build_sdist, choose_architectures]
-    name: Wheel for Linux ${{ matrix.cibw_arch }}
-    runs-on: ${{ matrix.os }}
+    needs: [build_sdist, choose_wheel_types]
+    name: ${{ matrix.wheel_type }} wheels
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
-        cibw_arch: ${{ fromJSON(needs.choose_architectures.outputs.cibw_arches) }}
+        wheel_type: ${{ fromJSON(needs.choose_wheel_types.outputs.wheel_types) }}
 
     steps:
       - name: Disable ptrace security restrictions
@@ -66,12 +67,12 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16.2
         env:
-          CIBW_BUILD: "cp3{7..12}-*"
-          CIBW_ARCHS_LINUX: ${{ matrix.cibw_arch }}
+          CIBW_BUILD: "cp3{7..12}-${{ matrix.wheel_type }}"
+          CIBW_ARCHS_LINUX: auto aarch64
           CIBW_PRERELEASE_PYTHONS: True
       - uses: actions/upload-artifact@v4
         with:
-          name: ${{ matrix.cibw_arch }}-wheels
+          name: ${{ matrix.wheel_type }}-wheels
           path: ./wheelhouse/*.whl
 
   test_attaching_to_old_interpreters:
@@ -85,7 +86,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/download-artifact@v4
         with:
-          name: "x86_64-wheels"
+          name: "manylinux_x86_64-wheels"
           path: dist
       - name: Set up Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


gusmonod:
@godlygeek this is now ready for review. Let me know what you think of this alternative approach.

.github/workflows/build_wheels.yml
```diff
@@ -35,20 +36,18 @@ jobs:
       - id: x86_64
         run: echo "cibw_arch=x86_64" >> $GITHUB_OUTPUT
       - id: aarch64
-        if: github.event_name == 'release' && github.event.action == 'published'
```
godlygeek, line 38: Restore this before landing this change

.github/workflows/build_wheels.yml
```diff
@@ -294,13 +294,13 @@ jobs:
   upload_pypi:
     needs: [test_wheels]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'published'
```
godlygeek, line 297: And this

.github/workflows/build_wheels.yml
```diff
@@ -294,13 +294,13 @@ jobs:
...
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          name: artifact
+          # with no name set, it downloads all of the artifacts
           path: dist
-      - uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          skip_existing: true
-          password: ${{ secrets.PYPI_PASSWORD }}
```
godlygeek, line 306: And this

godlygeek (changes requested):
Marking as "request changes" just so we don't accidentally land this before putting some of the "only on releases" checks back.

godlygeek:
Wow, OK.

First off, I think this looks a lot cleaner - both cleaner than #161, and also probably cleaner than what's on main.

It cuts the total amount of CPU time required for a full run (including aarch64 builds) down from 3.5 hours to 1.65 hours, which is borderline miraculous. It looks like the old approach was wasting a ton of time rebuilding `elfutils` from source over and over in a `before-all` step. This new approach only builds it once per architecture, which, yes, is what we ought to have done all along, and which saves an absolutely unbelievable amount of time when it comes to aarch64 builds. Those run under qemu, and are painfully slow, and the unnecessary repetitive builds of `elfutils` were using way more time than I realized.

It does mean that for a full run that includes the aarch64 builds, we can't test the x86-64 wheels until after the aarch64 ones have been built, which takes about an hour. That's a tad annoying, but oh well, that's why we don't build for aarch64 on every PR...

Building all the x86-64 wheels took 13 minutes, which means that on PRs there would be a 13 minute delay before the first tests start running. That's not _too_ terrible, but it looks like on main the wheels are are built in under 6 minutes thanks to the parallelism, so this would mean more than doubling the amount of time until we start to see test results. Maybe that's not so bad, though...

I think I'm comfortable landing this.

@pablogsal Any objections? The two downsides are
1. For releases, the tests won't run until after the aarch64 builds complete, which takes an hour.
2. For normal PRs, the tests will run as soon as all x86-64 wheels are built, but since we're sacrificing some parallelism, that will take a bit more than 2x as long as it used to.

In exchange for those tradeoffs, we reduce the total CPU time of a release run by more than half, we gain the ability to download the wheels while the tests are still running, and we get to upgrade from the v3 of `upload-artifact` and `download-artifact` to the v4.

I think that's a cost I'm willing to pay, but I want to make sure you're on board before landing this.

godlygeek:
For the x86-64 wheels, perhaps we should do the manylinux and musllinux separately. Doing those in parallel should get probably us back to running the tests as quickly as we used to...

godlygeek (approved):
Alright @gusmonod - I've pushed two new commits, which re-add the checks for steps that happen only on releases, and which switch musllinux and manylinux builds to happen in parallel. With that, the time to run the test suite on a normal (non-release) PR run is back down to about what it is on main, including the latency before the first tests start running.

Check my work on the two commits I pushed. If you're happy with them, I'm happy landing this.

.github/workflows/build_wheels.yml
```diff
@@ -24,32 +24,33 @@ jobs:
...
-    name: Decide which architectures to build wheels for
+  choose_wheel_types:
+    name: Decide which wheel types to build
     runs-on: ubuntu-latest
     steps:
-      - id: x86_64
-        run: echo "cibw_arch=x86_64" >> $GITHUB_OUTPUT
-      - id: aarch64
+      - id: manylinux_x86_64
+        run: echo "wheel_types=manylinux_x86_64" >> $GITHUB_OUTPUT
```
gusmonod, line 37: Why do you use an underscore (`_`) instead of a dash (`-`) for the separation between `manylinux` and `x86_64`? I would prefer a dash if possible, to make it clear that `x86_64` is really a single unit.

godlygeek: This gets expanded in `CIBW_BUILD: "cp3{7..12}-${{ matrix.wheel_type }}"` and cibuildwheel uses a `_` instead of a `-` there.  See https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip for examples, but we need that to expand to (for instance) `cp312-musllinux_x86_64`

.github/workflows/build_wheels.yml
```diff
@@ -24,32 +24,33 @@ jobs:
...
     outputs:
-      cibw_arches: ${{ toJSON(steps.*.outputs.cibw_arch) }}
+      wheel_types: ${{ toJSON(steps.*.outputs.wheel_types) }}
 
   build_wheels:
-    needs: [build_sdist, choose_architectures]
-    name: Wheel for Linux-${{ matrix.cibw_python }}-${{ matrix.cibw_arch }}
-    runs-on: ${{ matrix.os }}
+    needs: [build_sdist, choose_wheel_types]
+    name: ${{ matrix.wheel_type }} wheels
```
gusmonod, line 48: I would add `Build ` at the front there to make it clearer what this step does.

```suggestion
    name: Build ${{ matrix.wheel_type }} wheels
```

gusmonod (commented):
This all looks good to me! I have one nitpick below, feel free to ignore it if you want to. As I have created the PR myself though, I cannot approve it.

.github/workflows/build_wheels.yml
godlygeek: That gets truncated in the UI:

image

Compared to without:

image

I think it looks better without the word "build", and I think the intention is clear enough given that it's in a box labeled "Matrix: build_wheels", so I think I'm gonna skip this suggestion.

gusmonod: I guess it depends on what you prioritize, between the UI in the PR checks (where adding `Build` probably makes more sense) or the GH actions summary view (where it's better left off). Either way, this was just a nitpick, it is indeed easy enough to understand.

Issue #168, opened 2024-02-09
Support running on gz files directly


gaborbernat:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

Core dumps are often automatically archived to save disk space (in my case using the `.gz`) format. However, `pystack` does not allow running directly on these files, requiring a raw core dump. This means it introduces an extra manual step of unzipping the core dump before I can use pystack to see the stacktrace.

### Describe the solution you'd like

pystack should automate this kind of core dump extraction; it could see the file ends with `.gz` and call gzip unarchive in a temp file, and run against the temp file. This would be a quality of life improvement.

### Alternatives you considered

Keep doing the manual method...

godlygeek:
This sounds reasonable to me. It seems reasonable to check whether a file is or isn't a gzip file (not based on its file extension but based on its contents), and if so to extract it to a temp file. Without checking the code, I'm guessing that we probably need a `NamedTemporaryFile` and can't get by with just a `tempfile.TemporaryFile`

pablogsal:
Fixed by #171

PR #171, opened 2024-05-07
Add gzip support

ecalifornica:
*Issue number of the reported bug or feature request: #168 *

**Describe your changes**
Add check for core dump file ending in `.gz`. If so, decompress to temporary file and run analyzer on temporary file.

**Testing performed**
Added a unit test following the pattern of current tests. It mocks calls to `gzip` and `tempfile`, asserts that the analyzer was called with the correct path. It does not test for any failure modes, and really only tests the mocks. Happy to add more tests if requested.

**Additional context**

### DIFF
--- a/src/pystack/__main__.py
+++ b/src/pystack/__main__.py
@@ -1,11 +1,9 @@
 import argparse
-import gzip
 import logging
 import os
 import pathlib
 import signal
 import sys
-import tempfile
 from contextlib import suppress
 from textwrap import dedent
 from typing import Any
@@ -15,7 +13,9 @@ from typing import Optional
 from typing import Set
 
 from pystack.errors import InvalidPythonProcess
+from pystack.process import decompress_gzip
 from pystack.process import is_elf
+from pystack.process import is_gzip
 
 from . import errors
 from . import print_thread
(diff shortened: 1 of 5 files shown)


src/pystack/__main__.py
```diff
@@ -319,6 +321,19 @@ def process_core(parser: argparse.ArgumentParser, args: argparse.Namespace) -> N
     if not corefile.exists():
         parser.error(f"Core {corefile} does not exist")
 
+    temp_file = None
+    if corefile.suffix == ".gz":
```
pablogsal, line 325: Maybe we can improve this to not rely on the filename by checking the gz magic number: `1f 8b` from the header

pablogsal:
Thanks a lot for the PR! I will try to review it this week :)

src/pystack/__main__.py
ecalifornica: In making this change I hit a small snag: other tests mock the `pathlib.Path().exists()` check, causing them to fail when attempting to open the non-existent corefile in order to check the header. What do you think the correct approach would be here?
- Update those tests to expect the read error?
- Create a corefile fixture?
- Something else?

pablogsal: Move this code to a function and mock/patch it in those tests to return None or whatever indicates that this operation doesn't make sense

ecalifornica: Good call, I should've done this to begin with. Thanks!

pablogsal:
You  need a news entry (the other error in the Lint step you can ignore as we are fixing this in other PR).

ecalifornica:
Thanks for running the checks yesterday @godlygeek. I've picked up your fix for the Alpine test failure and removed the code that was causing the Python3.7 failure. Hopefully those changes should cause this PR to go green.

pablogsal (approved):

PR #175, opened 2024-05-13
Enable debuginfod support when testing wheels in Arch Linux

pablogsal:
Arch is more aggressively stripping out binaries and the tests that rely
on debug symbols fail due to insufficient debugging information. Getting
these debugging symbols by hand is a bit of a mess because Arch doesn't
publish debug packages in the same index. On the other hand Arch
provides a debuginfod server, so we can leverage that when running the
tests.

**Describe your changes**

**Testing performed**

**Additional context**


godlygeek (approved):
LGTM

PR #177, opened 2024-05-20
ci: Generate documentation on every push to main

gusmonod:
This helps having the changes to documentation be visible faster, which
is especially useful when we make rapid and frequent changes to it.


godlygeek (approved):
LGTM

Issue #53, opened 2023-04-23
Add a tutorial to the documentation


godlygeek:
Make our documentation more approachable by adding a quick tutorial showing how to get started with the basic features of PyStack.

manasomali:
I would like to work in this issue.

gusmonod:
Hi @Gigi1111 as part of the PyCon US 2024 sprints we are working on PyStack, and it's been a year since you said you wanted to work on this, so I'll assign it to someone else.

PR #179, opened 2024-05-20
docs: Add a tutorial for `pystack remote` - deadlock example

manasomali:
*Issue number of the reported bug or feature request: #53*

**Describe your changes**
Add the first tutorial about how to use PyStack to profiling the deadlock problem.

**Testing performed**
No tests added.

**Additional context**
Made during pycon2024 sprint. First day.

### DIFF
--- a/docs/tutorials/deadlock.py
+++ b/docs/tutorials/deadlock.py
@@ -1,0 +1,30 @@
+import os
+import threading
+import time
+
+
+def background(first_lock, second_lock):
+    with first_lock:
+        print("    First lock acquired")
+        time.sleep(1)
+        with second_lock:
+            print("    Second lock acquired")
+
+
+if __name__ == "__main__":
+    print(f"Process ID: {os.getpid()}")
+    lock_a = threading.Lock()
+    lock_b = threading.Lock()
+
+    t1 = threading.Thread(target=background, args=(lock_a, lock_b))
+    t2 = threading.Thread(target=background, args=(lock_b, lock_a))
+
+    print("Starting First Thread")
+    t1.start()
+    print("Starting Second Thread")
+    t2.start()
+
+    t1.join()
+    t2.join()
+
+    print("Finished execution")
--- a/docs/tutorials/deadlock.rst
+++ b/docs/tutorials/deadlock.rst
@@ -1,0 +1,237 @@
+Deadlock
+========
+
+Intro
+-----
+
+This lesson is meant to familiarize you with PyStack with a classic problem: lock acquisition.
+
+In this exercise, we will intentionally create a lock ordering issue, which is a common way of
+causing a deadlock, where two or more threads are all waiting for the others to release resources,
+causing the program to hang indefinitely.
+
+Development Environment Setup
+-----------------------------
+
+Navigate to the `PyStack GitHub repo <https://github.com/bloomberg/pystack>`_ and get a copy of the
+source code. You can either clone it, or just download the zip, whatever is your preference here.
+
+You will also need a terminal with a reasonably recent version of ``python3`` installed.
+
+Once you have the repo ready, ``cd`` into the ``docs/tutorials`` folder:
+
+.. code:: shell
+
+    cd docs/tutorials/
+
+It is here where we will be running the tests and exercises for the remainder of the tutorial.
+
+Let's go ahead and setup our virtual environment. For reference, here are the official `python3 venv
+docs <https://docs.python.org/3/library/venv.html>`_. You can also just follow along with the
+commands below.
+
+.. code:: shell
+
+    python3 -m venv .venv
+
+Once your virtual environment has been created, you can activate it like so:
+
+.. code:: shell
+
+    source .venv/bin/activate
+
+Your terminal prompt will be prefixed with ``(.venv)`` to show that activation was successful.
+With our virtual environment ready, we can go ahead and install PyStack:
+
+.. code:: shell
+
+    python3 -m pip install pystack
+
+Keep your virtual environment activated for the rest of the tutorial, and you should be able to run
+any of the commands in the exercises.
+
+Debugging a running process
+---------------------------
+
+``pystack remote`` lets you analyze the status of a running ("remote") process.
+
+Triggering the deadlock
+^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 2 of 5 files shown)


manasomali:
Sorry, fixing the lint.

docs/tutorials/deadlock.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env
+
+import os
+import threading
+import time
+
+
+def process1(lock1, lock2):
+    print("Process 1: Trying to acquire lock1")
+    lock1.acquire()
```
gusmonod, line 10: Can you please use the context manager instead of the `acquire` method? Something like:

```
print("Thread 1: Trying to acquire lock1")
with lock1:
    print("Thread 1: lock1 acquired")

    time.sleep(1)

    print("Thread 1: Trying to acquire lock2")
    with lock2:
        print("Thread 1: lock2 acquired")
print("Thread 1: Released both locks")
```

I think this makes the code a bit more explicit and probably removes the need for comments.

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
+Deadlock
+===============================
+
+Intro
+---------
```
gusmonod, line 5: Small change: I think this is a little more tidy this way.

```suggestion
Deadlock
========

Intro
-----
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code:: shell
+
+    python3 -m pip install -r requirements-tutorial.txt
+
+Keep your virtual environment activated for the rest of the tutorial, and you should be able to run
+any of the commands in the exercises.
+
+PyStack remote use
+------------------------
```
gusmonod, line 54: Same as above, a little more tidy this way

```suggestion
PyStack remote use
------------------
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
+Deadlock
+===============================
+
+Intro
+---------
+
+This lesson is meant to familiarize you with PyStack with a classical problem of locks acquisition.
+
+We will intentionally perform a deadlock, it a classic concurrency problem where improper ordering of lock acquisition can lead to deadlock,
+a situation where two or more threads are waiting for each other to release resources, causing the program to hang indefinitely.
```
gusmonod, line 10: I would suggest a slight rewording there:

```suggestion
In this exercise, we will intentionally create a lock ordering issue,
which is a common way of causing a deadlock, where two or more threads
are waiting for each other to release resources, causing the program to
hang indefinitely.
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    python3 -m pip install -r requirements-tutorial.txt
+
+Keep your virtual environment activated for the rest of the tutorial, and you should be able to run
+any of the commands in the exercises.
+
+PyStack remote use
+------------------------
+
+PyStack remote has the ability analyze the status of a running (remote) process.
```
gusmonod, line 56: typo

```suggestion
PyStack remote has the ability to analyze the status of a running (remote) process.
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+.. code:: shell
+
+    python3 deadlock.py
+
+This script will intentionally perform a deadlock and the first message will contain the process ID.
+Another option is to find it with:
+
+.. code:: shell
+
+    ps | grep deadlock.py
```
gusmonod, line 72: I would add these options here to make it more likely that there is a match.

```suggestion
    ps aux | grep deadlock.py
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+
+After the deadlock occurs we can use the PyStack command to analyze the process (replace the pid with the number of the process):
+
+.. code:: shell
+
+    pystack remote pid
```
gusmonod, line 78: In order to make it a bit clearer what needs to be done, using `<PID>` makes the command fail immediately if it isn't replaced.

```suggestion
After the deadlock occurs we can use the PyStack command to analyze the
process (replace <PID> with the process ID from the previous step):

.. code:: shell

    pystack remote <PID>
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+
+
+Potentially it will appear an `Operation not permitted` error. If it is your case use this command instead:
+
+.. code:: shell
+
+    sudo -E pystack remote pid
```
gusmonod, line 85: Same as above

```suggestion
    sudo -E pystack remote <PID>
```

docs/tutorials/deadlock.rst
```diff
@@ -0,0 +1,92 @@
...
+
+
+Understanding the results
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Notice that each section is displaying a running thread and linking the exactly line that is related to it.
+This way we can successful diagnose that there are a improper ordering of lock acquisition.
```
gusmonod, line 92: I think we could expand on that. We could actually print the expected output from PyStack and explains what it means, and how it shows that there is a lock ordering problem.

gusmonod (changes requested):
First of all, thanks for working on this! This is fantastic and a really helpful addition. I think there are a few things that can be improved though, that I have detailed below.

manasomali:
@gusmonod I don't know how to solve the error from build docs step. 🙃

manasomali:
PR updated, added a new section related to the use of `--locals`. Along with the recommended improvements in the deadlock.py script.

manasomali:
The test_in_alpine step is failing, what is the solution for this issue? 🤔

godlygeek:
> The test_in_alpine step is failing, what is the solution for this issue? 🤔

That's fixed by #185 

I'll pick this PR up soon and move it forward.

godlygeek (approved):
> I'll pick this PR up soon and move it forward.

"soon" 😅 

I've slightly tweaked some of the wording here and there, and dropped the `requirements-tutorial.txt` in favor of just doing a `python3 -m pip install pystack`, at least for now. Otherwise, this LGTM!

Thanks so much for the contribution, @manasomali!

PR #180, opened 2024-05-20
docs: Add a tutorial for `pystack remote` - prime example

gwen-sarapata:


godlygeek (approved):
Looks great to me! I've made some small tweaks to wording here and there, and restructured the layout of the tutorial to match the layout of the deadlock example (in particular, I dropped the solution file - it didn't make much sense to have it, since it only contains a one-line change, and it's easy to forget that we've got two copies and only modify one. They were already out of sync when I started reviewing!)

I kept all of the meat of this example the same, though, and I think it's a wonderful example full of misdirection that leads people away from the actual bug.

Thanks for the contribution, @gwen-sarapata!

PR #181, opened 2024-05-21
docs: Add a tutorial for `pystack core`

PistachioCannoli:
The tutorial walks users through a core dump configuration, then uses a python script to invoke a dump with stdlib SIGABRT, and then utilizes pystack to trace the error.

*Issue number of the reported bug or feature request: #53 *

**Describe your changes**
Wrote a tutorial page which guides users through a stack trace with pystack's core subcommand. The tutorial starts by introducing the concept, then configures the core dump settings such as size and output location. Afterwards, it uses a test script written which follows a series of asserts, if an assert is thrown, stdlib abort is called to interrupt the program, causing a core dump. Finally, the tutorial uses the core subcommand to trace the specific python stack then uses the --native flag to show the native C frames to assess the error.

**Testing performed**
Tested in an environment which copied the documentation structure of memray, but I reverted all memray dependencies prior to committing my changes.


godlygeek:
Thanks for the contribution, @PistachioCannoli! Sorry for taking so long to get around to reviewing it.

I decided that the example program that I had suggested to you was a bit too trivial to make for an entertaining tutorial, and after a bit of effort and a few conversations I managed to come up with a real way to crash the interpreter with a segfault that works on all Python versions Memray supports, and that's much more challenging to debug with PyStack than the trivial version that just calls `abort()` was. So, I've pushed a commit on top of yours that drops the example program we originally came up with in favor of this new one. I also renamed the files to fit into the structure established by the other tutorial PRs, and filled in a bit more information about how to control whether and how core files are written.

Thanks to all your work here, I had a great starting point.

@sarahmonod Since I modified this so heavily, I'd rather not be the one to land it. Can you take a look and ensure I haven't made any mistakes?

godlygeek:
(Ignore the failing DCO check; it's just mad that I edited the Signed-off-by line to use a real name instead of a pseudonym, but we have received the DCO)

pablogsal (approved):
LGTM

Issue #142, opened 2023-08-23
Consider a "brief" native mode only showing C frames below the last Python frame


godlygeek:
### Is there an existing proposal for this?

- [X] I have searched the existing proposals

### Is your feature request related to a problem?

We default to showing pure Python stacks, both for performance and for legibility to people who aren't systems programmers, but often these aren't very useful for figuring out exactly why a program had a hang or crash, since it's not always clear what might be happening inside of the current Python call. The solution to this is running with `--native` or `--native-all` mode, but these modes are both very verbose.

### Describe the solution you'd like

@alicederyn has pitched an idea for me: it might be possible to eliminate some of the noise and performance implications of `--native` mode while still providing enough information to puzzle out deadlocks or blocking happening in the C code called from the user's Python code if have a mode that only shows the C frames after the last Python frame. That is, instead of our pure-Python stacks:
```text
Traceback for thread 112 [] (most recent call last):
    (Python) File "/test.py", line 17, in <module>
        first_func()
    (Python) File "/test.py", line 6, in first_func
        second_func()
    (Python) File "/test.py", line 10, in second_func
        third_func()
    (Python) File "/test.py", line 14, in third_func
        time.sleep(1000)
```

And instead of the hybrid stacks that we show with `--native`:
```text
Traceback for thread 112 [] (most recent call last):
    (C) File "???", line 0, in _start ()
    (C) File "???", line 0, in __libc_start_main ()
    (C) File "Modules/main.c", line 743, in Py_BytesMain (/usr/lib/libpython3.8.so.1.0)
    (C) File "Modules/main.c", line 689, in Py_RunMain (/usr/lib/libpython3.8.so.1.0)
    (C) File "Modules/main.c", line 610, in pymain_run_python (inlined) (/usr/lib/libpython3.8.so.1.0)
    (C) File "Modules/main.c", line 385, in pymain_run_file (inlined) (/usr/lib/libpython3.8.so.1.0)
    (C) File "Python/pythonrun.c", line 472, in PyRun_SimpleFileExFlags (/usr/lib/libpython3.8.so.1.0)
    (C) File "Python/pythonrun.c", line 439, in pyrun_simple_file (inlined) (/usr/lib/libpython3.8.so.1.0)
    (C) File "Python/pythonrun.c", line 1085, in pyrun_file (/usr/lib/libpython3.8.so.1.0)
[...]

PR #182, opened 2024-05-21
Create native last mode

Kretzer6019:
*Issue number of the reported bug or feature request: #142*

**Describe your changes**
Created a new arg called `--native-last`, which shows only C frames after the last printed python frame. This helps to reduce verbose when trying to puzzle out deadlocks and blocking happening in the C code, which the last frames might be enouth.

**Testing performed**
Added `test_traceback_formatter_only_last_native_frames` to ensure that the new mode will show only the last C frames after the last python printed frame.

**Additional context**

### DIFF
--- a/src/pystack/__main__.py
+++ b/src/pystack/__main__.py
@@ -161,6 +161,15 @@ def generate_cli_parser() -> argparse.ArgumentParser:
         "the interpreter (implies --native)",
     )
     remote_parser.add_argument(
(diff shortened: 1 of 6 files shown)


godlygeek (approved):
LGTM. Sorry for the long delay on getting this landed, but this looks great and I'm landing it now. Thanks @Kretzer6019!

PR #184, opened 2024-05-24
Fallback to the first executable map if the given executable is invalid

pablogsal:
When a core file is created from a Python script that's executed
directly via shebang, the reported executable will be the shell script.
This is a problem because when we proceed to analyse the core file
pystack fails as this is not the correct binary that was used to
generate the core.

To avoid this problem, detect when this happens and fall back to the
executable reported in the first map that has a path in the core, which
is generally the real executable that we need.

**Describe your changes**

**Testing performed**

**Additional context**


src/pystack/__main__.py
```diff
@@ -357,9 +357,24 @@ def process_core(parser: argparse.ArgumentParser, args: argparse.Namespace) -> N
         print(format_failureinfo_information(corefile_analyzer.extract_failure_info()))
 
     if not is_elf(executable):
-        raise errors.InvalidExecutable(
-            f"The provided executable ({executable}) doesn't have an executable format"
+        first_map = next(
```
godlygeek, line 360: Should we maybe guard this with `if args.executable is not None:` to avoid falling back to the first file mapping if the user provided an executable on the command line and it's not valid?

It makes sense to me that, if our attempt to extract the executable from the core file gave us a non-ELF file, we should try again. It doesn't make sense to me for us to override the explicit choice provided by the user.

src/pystack/__main__.py
```diff
@@ -357,9 +357,24 @@ def process_core(parser: argparse.ArgumentParser, args: argparse.Namespace) -> N
...
+            (map for map in corefile_analyzer.extract_maps() if map.path is not None),
+            None,
         )
+        if (
+            first_map is not None
+            and first_map.path is not None
```
godlygeek, line 366: This condition is unnecessary, right? It's already checked in the genexpr passed to next()

godlygeek (changes requested):

src/pystack/__main__.py
pablogsal: Yeah, but is so mypy doesn't complain on the next call

src/pystack/__main__.py
pablogsal: Makes sense to me

godlygeek (approved):
You accidentally implemented the opposite of what I asked for - to fall back onto this autodetection only if the user _did_ provide an executable on the command line, instead of only if they _didn't_.

I've fixed that, and moved this detection up a bit higher in the file where it seems to fit better. After those changes, this LGTM. If you're happy with my changes, please squash them and 🚀

PR #185, opened 2024-05-24
ci: Fix test suite on Alpine

godlygeek:
We're no longer able to correctly identify `_PyEval_EvalFrameDefault` on Alpine without debug info.

This test was already adding symlinks to the split debug info for `ld-musl`, so let's extend that hack to also symlink to the split debug info for `libpython`.

### DIFF
--- a/tests/integration/test_relocatable_cores.py
+++ b/tests/integration/test_relocatable_cores.py
@@ -52,9 +52,14 @@ def test_single_thread_stack_for_relocated_core(
 
         # Hack: help libdw find the split debug for ld-musl on Alpine.
         # Otherwise, the test fails even though we do everything right.
-        ld_musl_debug = list(Path("/usr/lib/debug/lib/").glob("ld-musl*.debug"))
-        if ld_musl_debug != []:  # pragma: no cover
-            (relocated_bundle / "app" / ".debug").symlink_to("/usr/lib/debug/lib")
+        debuginfos = list(Path("/usr/lib/debug/lib/").glob("ld-musl*.debug"))
+        for debuginfo in debuginfos:  # pragma: no cover
+            (relocated_bundle / "app" / debuginfo.name).symlink_to(debuginfo)
+
+        # Apply the same hack to find the libpython debug info, too.
+        debuginfos = list(Path("/usr/lib/debug/usr/lib/").glob("libpython*.debug"))
+        for debuginfo in debuginfos:  # pragma: no cover
+            (relocated_bundle / "app" / debuginfo.name).symlink_to(debuginfo)
 
         binary_dependencies = ":".join(
             {str(file.parent) for file in relocated_bundle.glob("**/*.so*")}


godlygeek:
The native frames that we get without the libpython debug info are:
```text
(Pdb) pp thread.native_frames
[NativeFrame(address=139924869601034,
             symbol='libc_start_main_stage2',
             path='src/env/__libc_start_main.c',
             linenumber=95,
             colnumber=2,
             library='ld-musl-x86_64.so.1'),
 NativeFrame(address=139924865323059,
             symbol='PyEval_EvalCode',                                                                                                                                                                                                                                                         path='???',                                                                                                                                                                                                                                                                       linenumber=0,                                                                                                                                                                                                                                                                     colnumber=0,
             library='libpython3.12.so.1.0'),
 NativeFrame(address=139924864862184,
             symbol='PyObject_Vectorcall',
             path='???',
             linenumber=0,
             colnumber=0,
             library='libpython3.12.so.1.0'),
 NativeFrame(address=139924869877633,
             symbol='clock_nanosleep',
             path='src/time/clock_nanosleep.c',
             linenumber=34,
             colnumber=10,
             library='ld-musl-x86_64.so.1'),
 NativeFrame(address=139924869858839,
             symbol='__syscall_cp_c',
             path='src/thread/pthread_cancel.c',
             linenumber=33,
             colnumber=6,
             library='ld-musl-x86_64.so.1'),
 NativeFrame(address=139924869872018,                                                                                                                                                                                                                                                          symbol='__cp_end',                                                                                                                                                                                                                                                                path='src/thread/x86_64/syscall_cp.s',                                                                                                                                                                                                                                            linenumber=29,
             colnumber=0,
             library='ld-musl-x86_64.so.1')]
```

We see `PyObject_Vectorcall` and `PyEval_EvalCode`, but not `_PyEval_EvalFrameDefault`, so
```py
def _is_eval_frame(symbol: str, python_version: Tuple[int, int]) -> bool:
    if python_version < (3, 6):
        return "PyEval_EvalFrameEx" in symbol
    return "_PyEval_EvalFrameDefault" in symbol
```
is returning false for every frame, and we're not able to produce a hybrid stack.

tests/integration/test_relocatable_cores.py
```diff
@@ -53,8 +53,13 @@ def test_single_thread_stack_for_relocated_core(
         # Hack: help libdw find the split debug for ld-musl on Alpine.
         # Otherwise, the test fails even though we do everything right.
         ld_musl_debug = list(Path("/usr/lib/debug/lib/").glob("ld-musl*.debug"))
-        if ld_musl_debug != []:  # pragma: no cover
-            (relocated_bundle / "app" / ".debug").symlink_to("/usr/lib/debug/lib")
+        for debuginfo in ld_musl_debug:  # pragma: no cover
+            (relocated_bundle / "app" / debuginfo.name).symlink_to(debuginfo)
+
+        # Apply the same hack to find the libpython debug info, too.
+        ld_musl_debug = list(Path("/usr/lib/debug/usr/lib/").glob("libpython*.debug"))
```
pablogsal, line 60: Probably we want to rename this variable

pablogsal (approved):
LGTM with one nit :)

tests/integration/test_relocatable_cores.py
godlygeek: Oops, yep.

PR #186, opened 2024-06-04
Add support for Python 3.13

pablogsal:


src/pystack/_pystack/process.cpp
```diff
@@ -535,6 +535,20 @@ AbstractProcessManager::Pid() const
...
+AbstractProcessManager::getAddressFromCache(const std::string& symbol) const
+{
+    return d_symbol_cache[symbol];
+}
+
+void
+AbstractProcessManager::registerAddressInCache(const std::string& symbol, remote_addr_t address) const
+{
+    d_symbol_cache[symbol] = address;
+    return;
```
godlygeek, line 549:
```suggestion
```

src/pystack/_pystack/process.cpp
```diff
@@ -535,6 +535,20 @@ AbstractProcessManager::Pid() const
 {
     return d_pid;
 }
+
+remote_addr_t
+AbstractProcessManager::getAddressFromCache(const std::string& symbol) const
+{
+    return d_symbol_cache[symbol];
```
godlygeek, line 542: Er, I think you meant to do
```suggestion
    return d_type_cache[symbol];
```

src/pystack/_pystack/process.cpp
```diff
@@ -535,6 +535,20 @@ AbstractProcessManager::Pid() const
...
+}
+
+void
+AbstractProcessManager::registerAddressInCache(const std::string& symbol, remote_addr_t address) const
+{
+    d_symbol_cache[symbol] = address;
```
godlygeek, line 548: Er, I think you meant to do
```suggestion
    d_type_cache[symbol] = address;
```

src/pystack/_pystack/pyframe.cpp
```diff
@@ -62,7 +62,12 @@ FrameObject::getCode(
                << py_code_addr;
 
     uintptr_t last_instruction;
-    if (manager->versionIsAtLeast(3, 11)) {
+    if (manager->versionIsAtLeast(3, 13)) {
+        // In 3.13, the field is really called *instr_ptr*, but we are reusing the same
+        // field name as in 3.12 to avoid having to change the code in multiple places.
```
godlygeek, line 67: This doesn't explain the most surprising bit:
```suggestion
        // In 3.13 the field is called *instr_ptr* and points to the current instruction
        // rather than the previous one. We're reusing the same name and adjusting the
        // value to allow us to use the same decoding code across versions.
```

godlygeek (changes requested):

godlygeek (approved):
LGTM!

PR #194, opened 2024-07-31
Harden `--locals` against trashed memory

godlygeek:
When attempting to produce a repr for local variables, attempt to cope
better with trashed memory:

- Use "<invalid object at 0x1234>" as the repr of a `PyObject*` pointing
  directly at invalid memory, or one whose `ob_type` points at invalid
  memory.
- Use "<list object at 0x1234>" as, for instance, the repr of a list containing an
  invalid `ob_items` pointer. More generally, use this for any object
  where we can determine the type, but generating the repr according to
  that type's rules fails due to accessing unmapped memory.

Test this with some ctypes abuse.

### DIFF
--- a/src/pystack/_pystack/pytypes.cpp
+++ b/src/pystack/_pystack/pytypes.cpp
@@ -503,7 +503,7 @@ std::string
 GenericObject::toString(ssize_t max_size) const
 {
     std::stringstream os;
-    os << "<" << std::hex << std::showbase << d_classname << " at " << d_addr << ">";
+    os << "<" << std::hex << d_classname << " at 0x" << d_addr << ">";
     std::string object_str = os.str();
     return limitOutput(object_str, max_size);
 }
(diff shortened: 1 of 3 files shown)


src/pystack/_pystack/pytypes.cpp
```diff
@@ -503,7 +503,7 @@ std::string
 GenericObject::toString(ssize_t max_size) const
 {
     std::stringstream os;
-    os << "<" << std::hex << std::showbase << d_classname << " at " << d_addr << ">";
+    os << "<" << std::hex << d_classname << " at 0x" << d_addr << ">";
```
godlygeek, line 506: This change looks dumb, but it's because `std::showbase` doesn't show `0x` on 0, and `<invalid object at 0>` is more likely to confuse users than `<invalid object at 0x0>` because the latter is recognizable as an address.

tests/integration/test_local_variables.py
```diff
@@ -558,3 +558,71 @@ def second_func(the_argument):
...
+def test_trashed_locals(generate_threads, python, tmpdir):
+    # GIVEN
+    _, python_executable = python
+
+    program_template = """
+import ctypes
+import sys
+import time
+
+class ListObject(ctypes.Structure):
```
pablogsal, line 574: I am a bit hesitant because this test may change if we change/optimise the internal representation of these objects, which is not likely but also it's not extremely unlikely.

pablogsal (approved):
LGTM but left a comment regarding the (possible) flakiness of the test

tests/integration/test_local_variables.py
godlygeek: I considered that, but we already have [src/pystack/_pystack/cpython/list.h](https://github.com/bloomberg/pystack/blob/456f12bb9464e5dc38026fda0d9ed4a0271de019/src/pystack/_pystack/cpython/list.h) and [src/pystack/_pystack/cpython/tuple.h](https://github.com/bloomberg/pystack/blob/456f12bb9464e5dc38026fda0d9ed4a0271de019/src/pystack/_pystack/cpython/tuple.h) which will already need to be updated if CPython ever changes or if we ever start to support builds with `-DPy_TRACE_REFS`. This just adds one more place we'll need to update, but it's no more fragile than what the tool itself already does.

godlygeek: In any event, if we ever need to, we can test this with a custom C extension that trashes types that we own and can control the layout of. ctypes is a quick and dirty way to do it from pure Python, but these changes aren't specific to `list` and `tuple`, and we could instead test them with our own extension types with a bit more work if we ever have a reason to.

PR #195, opened 2024-08-03
Publish wheels for Python 3.13

godlygeek:
Now that Python 3.13 is ABI stable, document that it is supported and begin to publish wheels for it.

We haven't cut a release since we added support for 3.13, so we don't need another news entry. The existing one simply says "Add support for Python 3.13".

### DIFF
--- a/.github/ISSUE_TEMPLATE/---bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/---bug-report.yaml
@@ -53,6 +53,7 @@ body:
         - "3.10"
         - "3.11"
         - "3.12"
+        - "3.13"
     validations:
       required: true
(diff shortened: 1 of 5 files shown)


godlygeek:
Hm - should we document in the news entry that this has not been tested (and almost certainly doesn't work with) free threaded builds?

pablogsal (approved):

pablogsal:
> Hm - should we document in the news entry that this has not been tested with (and almost certainly doesn't work with) free threaded builds?

If we do this better to do in a separate PR. We may want to document this in the docs somewhere

pablogsal:
We may want to modify PyStack to refuse to work over free threaded builds

PR #197, opened 2024-08-07
Automatically check our offsets against CPython's self-reported debug offsets

godlygeek:
And fix two mistakes that were uncovered by this!


pablogsal (approved):

pablogsal:
Man, so glad we this this ❤️

PR #202, opened 2024-08-09
Fix compiler errors and some error paths

pablogsal:
When compiling pystack with gcc 14, the C++ standard library headers are
not giving us transitevely the 'algorithm' include and we are missing it
explicitly so it currently fails to compile.

Additionally, we are not properly handling some of the exceptions that
are being raised when something fails when copying memory when we are
resolving the Python version.

**Describe your changes**

**Testing performed**

**Additional context**

### DIFF
--- a/src/pystack/_pystack/mem.cpp
+++ b/src/pystack/_pystack/mem.cpp
@@ -368,7 +368,7 @@ CorefileRemoteMemoryManager::copyMemoryFromProcess(remote_addr_t addr, size_t si
     StatusCode ret = getMemoryLocationFromCore(addr, &offset_in_file);
 
     if (ret == StatusCode::SUCCESS) {
-        if (static_cast<size_t>(offset_in_file) > d_corefile_size) {
+        if (size > d_corefile_size || static_cast<size_t>(offset_in_file) > d_corefile_size - size) {
             throw InvalidRemoteAddress();
         }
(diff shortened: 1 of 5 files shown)


src/pystack/_pystack/mem.cpp
```diff
@@ -368,7 +368,7 @@ CorefileRemoteMemoryManager::copyMemoryFromProcess(remote_addr_t addr, size_t si
     StatusCode ret = getMemoryLocationFromCore(addr, &offset_in_file);
 
     if (ret == StatusCode::SUCCESS) {
-        if (static_cast<size_t>(offset_in_file) > d_corefile_size) {
+        if (static_cast<size_t>(offset_in_file + size) > d_corefile_size) {
```
godlygeek, line 371: This should be written a bit differently to avoid integer overflow bugs, I think. It's a bit pedantic, but I think we should do:
```suggestion
        if (size > d_corefile_size || static_cast<size_t>(offset_in_file) > d_corefile_size - size) {
```

I'm sure this doesn't matter at all for 64 bit builds, but it might for 32 bit builds.

godlygeek (changes requested):
One minor nit:

godlygeek (approved):
Discussed offline with Pablo, made that change, and rebased, after which this LGTM

Issue #199, opened 2024-08-09
CPython assertion failures cause odd behavior


ZeroIntensity:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Running with Python 3.12.5 (unreleased) compiled with `--with-pydebug` and `--with-assertions` seems to cause some odd behavior with PyStack, when the core file fails at an assertion.

Earlier today, I noticed PyStack segfaulted on a core file that did this (although, I'm struggling to reproduce this again), and I also observed it to have some highlighting issues (but not segfault) on a different case (but still failing at an assertion).

I've emailed a core file containing the incorrect highlight to @godlygeek, but I'm still working on coming up with a reproducer for the segfault. 

### Expected Behavior

It shouldn't do that :stuck_out_tongue:

### Steps To Reproduce

1. Clone the 3.12 branch of CPython
2. `./configure --with-pydebug --with-assertions`
3. Add some dummy failed assertion somewhere. You could make a new function, add to an existing one, the choice is yours -- it just needs to fail and cause a coredump.
4. Run with `pystack core --native-all`

### Pystack Version

1.3.0

### Python Version

3.12

### Linux distribution

Arch Linux

pablogsal:
I cannot reproduce this:

```
$ ./python -m sysconfig | grep 'pydebug'
        CONFIG_ARGS = "'--with-pydebug' '--with-assertions'"
$ ./python --version
Python 3.12.5+
$ git --no-pager diff --no-color
diff --git a/Modules/_heapqmodule.c b/Modules/_heapqmodule.c
index 00285ae01f8..d35e674f8bb 100644
--- a/Modules/_heapqmodule.c
+++ b/Modules/_heapqmodule.c
@@ -146,6 +146,7 @@ heappop_internal(PyObject *heap, int siftup_func(PyListObject *, Py_ssize_t))

     /* raises IndexError if the heap is empty */
     n = PyList_GET_SIZE(heap);
+    assert(n == 0);
     if (n == 0) {
         PyErr_SetString(PyExc_IndexError, "index out of range");
         return NULL;

$ ./python -c 'import heapq; heapq.heappop([1,2,3])'
python: ./Modules/_heapqmodule.c:149: heappop_internal: Assertion `n == 0' failed.
[1]    1427571 IOT instruction (core dumped)  ./python -c 'import heapq; heapq.heappop([1,2,3])'

[...]

PR #204, opened 2024-08-10
Protect against shim frames at the bottom of the stack

pablogsal:
It's possible that pystack analyses a core file or a program at a
specific moment where the bottom of the stack has a shim frame but a
normal frame has not been pushed yet or the last normal frame has been
removed because the evaluation loop is returning.

In this case, we were not resolving the code object, but we are not
skipping the frame either because it's the first frame. This was causing
the code object pointer in the frame to be NULL but we are not checking
for that condition when unwinding.

Closes: #199

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -6,6 +6,7 @@ import pathlib
 import shutil
 import subprocess
 import sys
+import textwrap
 import time
 from typing import Generator
 from typing import Iterable
@@ -135,6 +136,38 @@ def generate_core_file(
             process.terminate()
             process.kill()
             process.wait(timeout=TIMEOUT)
+
+
+@contextlib.contextmanager
+def generate_core_after_crash(
+    python: pathlib.Path,
+    test_file: pathlib.Path,
+    tmpdir: pathlib.Path,
+) -> Generator[pathlib.Path, None, None]:
+    gdb_commands = textwrap.dedent(
+        f"""
+        set pagination off
+        run
+        generate-core-file {tmpdir}/core.%p
+        quit
+        """
+    )
+
+    gdb_script = tmpdir / "gdb_script.txt"
+    with open(gdb_script, "w") as f:
+        f.write(gdb_commands)
+
+    subprocess.run(
+        ["gdb", "-x", str(gdb_script), "-batch", "--args", str(python), str(test_file)],
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        timeout=TIMEOUT,
+        text=True,
+    )
+
+    (core,) = pathlib.Path(tmpdir).glob("core.*")
+    yield core
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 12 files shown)


pablogsal:
@godlygeek we still have a problem with shim frames and native frames in general because it's possible that we are in one of these two situations (we are just returning or we are just entering) and therefore we won't identify the entry frame properly as in 3.12+ entry frames are frames preceded by shim frames but out shim frame it's at the bottom of the stack and this prevents us from merging the native stack properly

pablogsal:
For instance, this generates such a situation:

```
import asyncio
import _asyncio
import gc
import time


class MyFuture(asyncio.Future):
    def __del__(self):
        time.sleep(1000)


async def main():
    it = iter(MyFuture())

asyncio.run(main())
```

pablogsal:
I decided to fix the shim problem by always capturing shim frames and filter them out properly in the traceback formatter, which is slightly less efficient because we are not skipping them but thanks to the cache the overhead is minimal because the code object is always the same.

tests/utils.py
```diff
@@ -268,3 +299,25 @@ def all_pystack_combinations(corefile=False, native=False):
     [python[:2] for python in AVAILABLE_PYTHONS if python.version < (3, 10)],
     ids=[python[1].name for python in AVAILABLE_PYTHONS if python.version < (3, 10)],
 )
+
+
+def all_pythons_between(
+    *, from_version: Tuple[int, int], to_version: Optional[Tuple[int, int]] = None
+):
```
godlygeek, line 306: This seems unnecessarily generic - you've defined a `all_pythons_between` that takes a mandatory start version and an optional end version, but you never provide that end version anywhere, so this feels like YAGNI. It seems like you could just do `all_pythons_since()` for now, and wait to define an `all_pythons_between` until we need it.

tests/utils.py
```diff
@@ -137,6 +138,36 @@ def generate_core_file(
             process.wait(timeout=TIMEOUT)
 
 
+@contextlib.contextmanager
+def generate_core_at_breakpoint(
```
godlygeek, line 142: This function's name no longer matches what it does, since you don't actually set any breakpoints. `generate_core_after_crash` or something would be more appropriate.

tests/utils.py
```diff
@@ -137,6 +138,36 @@ def generate_core_file(
...
+    python: pathlib.Path,
+    test_file: pathlib.Path,
+    tmpdir: pathlib.Path,
+) -> Generator[pathlib.Path, None, None]:
+    gdb_commands = f"""
+set pagination off
+run
+generate-core-file {tmpdir}/core.%p
+quit
+"""
```
godlygeek, line 152: nit, but I don't like seeing stuff flushed to the left column. Can we do:
```suggestion
    gdb_commands = f"""
        set pagination off
        run
        generate-core-file {tmpdir}/core.%p
        quit
    """
```
Or, if gdb isn't ok with leading whitespace, can we add a textwrap.dedent?

godlygeek (changes requested):
A few small requests

pablogsal:
I have amended the last commit with your requests, please take another look

godlygeek:
LGTM, though I made [two more tweaks to the tests](https://github.com/bloomberg/pystack/compare/c05979cfe0b95ba13b1258fb17e62e2352291d37..0575232f365210bb47b8c4b65713be1968fe7068):
1. changed `all_pythons_since` to take major/minor as two args rather than a 2-tuple, to make it easier to call
2. updated the indentation of the f-string in `generate_core_after_crash` to prevent the end triple quotes from being less indented than the start triple quotes (and dropped the `\` after the start triple quotes; gdb doesn't mind the leading blank line)

godlygeek (approved):

PR #206, opened 2024-08-16
Use _Py_DebugOffsets instead of hardcoded offsets when possible

godlygeek:
Python 3.13 adds a new `_Py_DebugOffsets` structure to `_PyRuntimeState` which tells us more or less everything that we need to know in order to decode the interpreter's data structures. Look for this structure and, if we find it, use the offsets contained in it rather than the offsets compiled into pystack at build time.

This allows us to decode stacks for interpreters with different build time sizes of different structures, including those running in a `python3.13t` free-threaded interpreter.

### DIFF
--- a/src/pystack/_pystack/structure.h
+++ b/src/pystack/_pystack/structure.h
@@ -1,0 +1,91 @@
+#pragma once
+
+#include <array>
+#include <vector>
+
+#include "process.h"
+
+namespace pystack {
+
+template<typename OffsetsStruct>
+class Structure
+{
+  public:
+    // Constructors
+    Structure(std::shared_ptr<const AbstractProcessManager> manager, remote_addr_t addr);
+    Structure(const Structure&) = delete;
+    Structure& operator=(const Structure&) = delete;
+
+    // Methods
+    void copyFromRemote();
+
+    template<typename FieldPointer>
+    remote_addr_t getFieldRemoteAddress(FieldPointer OffsetsStruct::*field) const;
+
+    template<typename FieldPointer>
+    const typename FieldPointer::Type& getField(FieldPointer OffsetsStruct::*field);
+
+  private:
+    // Data members
+    std::shared_ptr<const AbstractProcessManager> d_manager;
+    remote_addr_t d_addr;
+    ssize_t d_size;
+    std::array<char, 512> d_footprintbuf;
+    std::vector<char> d_heapbuf;
+    char* d_buf;
+};
+
+template<typename OffsetsStruct>
+inline Structure<OffsetsStruct>::Structure(
+        std::shared_ptr<const AbstractProcessManager> manager,
+        remote_addr_t addr)
+: d_manager(manager)
+, d_addr(addr)
+, d_size(d_manager->offsets().get<OffsetsStruct>().size)
+, d_buf{}
+{
+}
+
+template<typename OffsetsStruct>
+inline void
+Structure<OffsetsStruct>::copyFromRemote()
+{
+    if (d_buf) {
+        return;  // already copied
+    }
+
+    if (d_size < 512) {
+        d_buf = &d_footprintbuf[0];
+    } else {
+        d_heapbuf.resize(d_size);
+        d_buf = &d_heapbuf[0];
+    }
+    d_manager->copyMemoryFromProcess(d_addr, d_size, d_buf);
+}
+
+template<typename OffsetsStruct>
+template<typename FieldPointer>
+inline remote_addr_t
+Structure<OffsetsStruct>::getFieldRemoteAddress(FieldPointer OffsetsStruct::*field) const
+{
+    offset_t offset = (d_manager->offsets().get<OffsetsStruct>().*field).offset;
+    return d_addr + offset;
+}
+
+template<typename OffsetsStruct>
+template<typename FieldPointer>
+inline const typename FieldPointer::Type&
+Structure<OffsetsStruct>::getField(FieldPointer OffsetsStruct::*field)
+{
+    copyFromRemote();
+    offset_t offset = (d_manager->offsets().get<OffsetsStruct>().*field).offset;
+    if (d_size < 0 || (size_t)d_size < sizeof(typename FieldPointer::Type)
+        || d_size - sizeof(typename FieldPointer::Type) < offset)
+    {
+        abort();
+    }
+    auto address = d_buf + offset;
+    return *reinterpret_cast<const typename FieldPointer::Type*>(address);
+}
+
+}  // namespace pystack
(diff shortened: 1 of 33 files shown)


src/pystack/_pystack/version.h
```diff
@@ -138,13 +125,15 @@ struct py_thread_v
 
 struct py_runtime_v
 {
-    typedef PyRuntimeState Structure;
     ssize_t size;
     FieldOffset<remote_addr_t> o_finalizing;
     FieldOffset<remote_addr_t> o_interp_head;
-    FieldOffset<GCRuntimeState> o_gc;
+    FieldOffset<char> o_gc;
```
pablogsal, line 131: Hummm why `char`? Isn't this a pointer?

src/pystack/_pystack/structure.h
```diff
@@ -0,0 +1,85 @@
...
+    return d_addr + offset;
+}
+
+template<typename OffsetsStruct>
+template<typename FieldPointer>
+inline const typename FieldPointer::Type&
+Structure<OffsetsStruct>::getField(FieldPointer OffsetsStruct::*field)
+{
+    copyFromRemote();
+    offset_t offset = (d_manager->offsets().get<OffsetsStruct>().*field).offset;
```
pablogsal, line 80: Should we add an assert to bound check against `d_buf`?

tests/integration/test_gil.py
```diff
@@ -14,6 +17,15 @@
 TEST_SINGLE_THREAD_FILE = Path(__file__).parent / "single_thread_program.py"
 
 
+@pytest.fixture(autouse=True)
+def enable_gil_if_free_threading(python, monkeypatch):
+    _, python_executable = python
+    proc = subprocess.run([python_executable, "-Xgil=1", "-cpass"], capture_output=True)
```
pablogsal, line 23: This will always succeed as an invalid `-X` option doesn't raise or warn. I am missing something here?

tests/integration/test_local_variables.py
```diff
@@ -573,31 +573,33 @@ def test_trashed_locals(generate_threads, python, tmpdir):
...
 class TupleObject(ctypes.Structure):
     _fields_ = [
-        ("ob_refcnt", ctypes.c_ssize_t),
         ("ob_type", ctypes.c_void_p),
         ("ob_size", ctypes.c_ssize_t),
         ("ob_item0", ctypes.c_void_p),
         ("ob_item1", ctypes.c_void_p),
     ]
 
+def ob_type_field(obj):
```
pablogsal, line 589: Is this something we can avoid with the new offsets we added to 3.13?

pablogsal (approved):
LGTM modulo the things that we discussed offline

src/pystack/process.py
```diff
@@ -20,7 +20,8 @@
 )
 
 BSS_VERSION_REGEXP = re.compile(
-    rb"((2|3)\.(\d+)\.(\d{1,2}))((a|b|c|rc)\d{1,2})?\+? (\(.{1,64}\))"
+    rb"((2|3)\.(\d+)\.(\d{1,2}))((a|b|c|rc)\d{1,2})?\+?"
+    rb"(?: experimental free-threading build)? (\(.{1,64}\))"
```
pablogsal, line 24:
```suggestion
    rb"""
    (                   # Capturing group for the main version
        (2|3)           # Major version: 2 or 3
        \.              # Literal dot
        (\d+)           # Minor version: one or more digits
        \.              # Literal dot
        (\d{1,2})       # Patch version: 1 or 2 digits
    )
    (                   # Capturing group for pre-release identifier (optional)
        (a|b|c|rc)      # Pre-release type: alpha, beta, candidate, or release candidate
        \d{1,2}         # Pre-release version: 1 or 2 digits
    )?                  # The entire pre-release group is optional
    \+?                 # Optional plus sign for build metadata
    (?:                 # Non-capturing group for experimental build info (optional)
        \s              # Space
        experimental\ free-threading\ build
    )?
    \s                  # Space
    (                   # Capturing group for additional build info
        \(              # Opening parenthesis
        .{1,64}         # 1 to 64 characters of any type (except newline)
        \)              # Closing parenthesis
    )
    """,
    re.VERBOSE | re.BINARY
```

pablogsal: Matt doesn't like my suggestion :(

pablogsal: maybe add some examples:

```
# Sample strings that match the regex:
#
# b"3.8.10 (default, May 26 2023, 14:05:08)"
# b"2.7.18rc1 (v2.7.18rc1:8d21aa21f2, Apr 20 2020, 13:19:08)"
# b"3.9.5+ experimental free-threading build (Python)"
# b"3.10.0b2 (main, Jun 22 2021, 12:46:30)"
# b"2.7.15+ (default, Mar  5 2023, 16:38:35)"
```

src/pystack/_pystack/version.h
godlygeek: Yeah, but it's an inline structure whose layout we don't know at compile time, and so there's not a useful version-independent type we can use to refer to it by here. `char` is a hack to deal with the fact that we can't actually dereference this field into a meaningful `GCRuntimeState` directly, and instead need to take a pointer to it and create a `Structure` from that pointer.

I'll add a comment explaining this.

tests/integration/test_gil.py
godlygeek: An unrecognized `-X` option doesn't raise or warn, but `python3.13` recognizes `-Xgil` and fails when the build is not free-threading:
```
$ python3.13 -Xgil=1
Fatal Python error: config_read_gil: PYTHON_GIL / -X gil are not supported by this build
Python runtime state: preinitialized
```

tests/integration/test_local_variables.py
godlygeek: It's technically possible, but would require much more complexity. We'd need a way for pystack to attach to a running process and dump the offsets rather than the stack, and we'd need to then take those offsets and pass them to program being run by the test.

We've got options if this turns out to be a maintenance burden, but for now, I'd rather keep this as simple as we can.

tests/integration/test_gil.py
godlygeek: I've opened an upstream bug report for this annoying behavior, https://github.com/python/cpython/issues/123275

src/pystack/process.py
godlygeek: Comments shouldn't explain _what_ the code is doing, they should explain _why_ it's doing those things 😄 

To that end, the example strings are much more instructive than the plain English explanation of the regex syntax. I've taken 3 of them that seem to cover all the interesting cases.

godlygeek:
@pablogsal I've addressed all of your suggestions. I've also rebased this on main, and fixed conflicts that arose due to the fix for handling shim frames. Take another look when you get a chance, please.

pablogsal (approved):

PR #207, opened 2024-08-30
Support 3.13rc2

godlygeek:
New fields were added to `_Py_DebugOffsets` in 3.13rc2 which make make us unable to recognize the interpreter's data structures in memory. This PR contains two commits: one to update our static offsets to account for the newly added fields so that we can correctly attach to 3.13rc2 processes (at the cost that we're no longer able to attach to 3.13rc1 or earlier), and one to begin utilizing the new fields that were added to the offsets to remove assumptions that we were making in PyStack in favor of known correct answers from the interpreter.

Marking as draft for now because this will fail until 3.13rc2 is available.


godlygeek:
OK, this passes CI now. The problem wasn't cibuildwheel still using rc1: it still is, but that doesn't matter to us because we never run any tests with the Python version provided by cibuildwheel. We use it only for building, and it's fine to build our extension with 3.13rc1 as long as you only attach to processes using 3.13rc2 or higher. The problem was that actions/setup-python hadn't yet picked up rc2, but now it has, and so CI passed on a rerun.

I've marked this as `skip news`; the news entry added in #206 covers it:

> Support debugging free-threading (a.k.a. "nogil") Python 3.13 builds. Note that PyStack can't itself be run with `python3.13t`, it can only attach to a `python3.13t` process or core file from another interpreter.

godlygeek:
And, once we get this landed, we can cut a release. We don't expect any more changes to affect our 3.13 support from this point forward.

pablogsal (approved):

PR #209, opened 2024-09-11
Fix decoding Python 3.13 stacks without symbols

### LINKED ISSUE #205: Can't get even a pure-Python stack from a 3.13 without symbols or ELF files
These recently introduced changes:

https://github.com/bloomberg/pystack/blob/d6b11022345b1b429d9fc6fff311236fa3be2c53/src/pystack/_pystack/pycode.cpp#L168-L185

have left us unable to render even a pure-Python stack from a 3.13 core file unless either debug symbols or the original ELF files are available.

At the very least, we need to fail open rather than closed here: if we can't tell for certain whether anything is a code object, we'd do better to assume that everything is than that nothing else. But perhaps there's something better we can do.

godlygeek:
The eager `isValid` check that we implemented originally fails when symbols aren't available, as we're unable to look up `PyCode_Type` to compare the type of `f_executable` against it.

Instead, let's treat every `f_executable` as though it's a `PyCode_Type` without checking in advance. If it's not able to be interpreted as a `PyCode_Type`, we'll expect a `RemoteMemCopyError` from chasing an invalid pointer at some point, and can handle that the way that we have been handling `!isValid`.

Closes #205 

I tested this manually and confirmed that it does allow us to successfully decode a Python 3.13 core's Python stack even without the correct executable being available. There's no automated test added here, but in fairness the old `isValid` behavior that's being replaced was also not covered by any automated test.

### DIFF
--- a/src/pystack/_pystack/pycode.cpp
+++ b/src/pystack/_pystack/pycode.cpp
@@ -162,41 +162,11 @@ getLocationInfo(
     return location_info;
 }
 
-static bool
-isValid(const std::shared_ptr<const AbstractProcessManager>& manager, remote_addr_t addr)
-{
-    if (manager->versionIsAtLeast(3, 13)) {
-        // In Python 3.13, the frame f_executable field can be a code object or a bunch
-        // of other possible types (including None). We consider valid only the cases
-        // where it is a code object.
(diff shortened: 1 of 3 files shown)


godlygeek:
Adding "skip news" since this fixes a bug that never made it into a release.

pablogsal (approved):

PR #221, opened 2025-02-12
Handle failures to check if a file is ELF

godlygeek:
While attempting to find the executable for a core file, it's possible
that we extract the empty string from the core file where an executable
name ought to have been. In that case we construct a `pathlib.Path` with
the empty string, which happens to behave like a path of `.` - which is
a directory, not a file. This leads to failures in surprising places.

Since the core file may have been generated on a remote machine, or at
a distant point in time, we shouldn't assume that any file name found in
necessarily refers to a file that we can read just because it exists on
disk. It may have been replaced by a directory, or it may have its
permissions set up so that it's not readable by the user we're running
as. Rather than trusting that anything that we can `stat` must be a file
we can read, have our `is_elf` function treat any path from which we
fail to read as though it's not an ELF executable. This lets us continue
processing core files even if files they refer to can't be read by us,
letting us try fallbacks or otherwise attempt to make the best of
things.

### DIFF
--- a/news/221.bugfix.rst
+++ b/news/221.bugfix.rst
@@ -1,0 +1,1 @@
+Improve handling of core files where we cannot determine the executable.
--- a/src/pystack/__main__.py
+++ b/src/pystack/__main__.py
@@ -331,7 +331,7 @@ def process_core(parser: argparse.ArgumentParser, args: argparse.Namespace) -> N
     if args.executable is None:
         corefile_analyzer = CoreFileAnalyzer(corefile)
         executable = pathlib.Path(corefile_analyzer.extract_executable())
-        if not executable.exists() or not is_elf(executable):
+        if not is_elf(executable):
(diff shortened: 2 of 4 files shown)


godlygeek:
This will need to be rebased once #220 lands, since I've cherry picked over the two fixes needed to get CI running.

shollingsworth:
I just ran into this as well.

godlygeek:
There is almost certainly an issue with the core file, yeah. We shouldn't be extracting an empty executable string from it. You might get further if you pass the path to the executable that crashed as the optional last argument to `pystack remote` - so

```sh
/opt/venv/bin/pystack -vv core /path/to/core /path/to/executable
```

At the very least, that `-vv` should give you a bunch more debugging info that we could use to try to make sense of what's failing.

pablogsal (approved):

Issue #229, opened 2025-05-19
Add support for Python 3.14


godlygeek:
Update PyStack so that it's test suite successfully runs for Python 3.14 beta 1

dpinol:
On linux with python 3.14.2, pystack 1.6.0 fails
` Engine error: Invalid address in remote process 💀`

pablogsal:
I tested with `3.14.2` and it passes the test suite so there is something else at hand

PR #230, opened 2025-05-19
Add support for 3.14

ZeroIntensity:
*Issue number of the reported bug or feature request: #229*

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -323,8 +323,7 @@ jobs:
             python3-dev \
             python3-pip \
             python3-venv \
-            python3-dbg \
-            python3-distutils
+            python3-dbg
       - uses: actions/download-artifact@v4
         with:
           name: "manylinux_x86_64-wheels"
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -353,6 +353,7 @@ jobs:
           mv dist/sdist/*.tar.gz dist/
           mv dist/*-wheels/*.whl dist/
           rmdir dist/{sdist,*-wheels}
+          rm -f dist/*cp314*
           ls -R dist
       - uses: pypa/gh-action-pypi-publish@release/v1
         with:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.0.1
         env:
-          CIBW_BUILD: "cp3{8..14}-${{ matrix.wheel_type }}"
+          CIBW_BUILD: "cp3{8..14}{t,}-${{ matrix.wheel_type }}"
           CIBW_ARCHS_LINUX: auto aarch64
           CIBW_ENABLE: cpython-prerelease
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -353,7 +353,6 @@ jobs:
           mv dist/sdist/*.tar.gz dist/
           mv dist/*-wheels/*.whl dist/
           rmdir dist/{sdist,*-wheels}
-          rm -f dist/*cp314*
           ls -R dist
       - uses: pypa/gh-action-pypi-publish@release/v1
         with:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.0.1
         env:
-          CIBW_BUILD: "cp3{8..13}-${{ matrix.wheel_type }}"
+          CIBW_BUILD: "cp3{8..14}-${{ matrix.wheel_type }}"
           CIBW_ARCHS_LINUX: auto aarch64
           CIBW_ENABLE: cpython-prerelease
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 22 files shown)


godlygeek:
@ZeroIntensity I've managed to get CI working for 3.14 in #237, and rebased this PR on top of those changes. After getting CI running, there are quite a few failing tests here. Would you mind taking a look and trying to fix things?

At a glance, all of the failures seem to be related to reading local variables from Python 3.14 processes.

ZeroIntensity:
Will take a look soon.

godlygeek:
CI is using 3.14.0 beta 1

ZeroIntensity:
Sorry for the delay! I can't reproduce the CI failures locally on 3.14b1, but I *can* reproduce off CPython main, and changing the bits doesn't seem to fix it. How sure are we that it's 3.14b1?

ZeroIntensity:
Beta 2 is out anyway, I guess. Since I can repro on that, let's switch to it.

godlygeek:
> Beta 2 is out anyway

Indeed, if I rerun the failing CI job it now uses beta2 - seems like actions/setup-python now supports it.

And with beta2 there's a bunch more failures - 118 tests fail, instead of the previous 18 failures. So it definitely does seem like something changed between beta1 and beta2 that we'll need to adjust for.

godlygeek:
That must be https://github.com/python/cpython/pull/134286

ZeroIntensity:
Oh, yeah, that'd do it!

godlygeek:
https://github.com/python/cpython/commit/1822f33b1af989ebb9e7a5b58bd40721fca9a218 affected the debug offsets, too. Adding those gets us back down to 18 failing tests - probably the same ones that were failing with beta 1...

godlygeek:
I took another swing at dealing with the layout changes to `PyTupleObject`, and now everything is passing except for the tests running pystack against a 3.14t interpreter. That's pretty badly broken, it seems.

ZeroIntensity:
Which objects does pystack need to know the layout of? Free-threaded objects have a lot of internal layout changes.

godlygeek:
ints, floats, strings, bytes, dicts, lists, tuples, code objects, frames, and a handful of others.

The first issue I tracked down is that the `state` field of `PyASCIIObject` has a different layout in free-threaded. We'll need to update `AbstractProcessManager::getStringFromAddress` to cope with that, but we can get past it temporarily by just doing:

```diff
diff --git a/src/pystack/_pystack/process.cpp b/src/pystack/_pystack/process.cpp
index b48a225..8c3715c 100644
--- a/src/pystack/_pystack/process.cpp
+++ b/src/pystack/_pystack/process.cpp
@@ -540,10 +540,10 @@ AbstractProcessManager::getStringFromAddress(remote_addr_t addr) const
                    << addr;
         Structure<py_unicode_v> unicode(shared_from_this(), addr);

-        Python3::_PyUnicode_State state = unicode.getField(&py_unicode_v::o_state);
-        if (state.kind != 1 || state.compact != 1) {
-            throw InvalidRemoteObject();
-        }
+        //Python3::_PyUnicode_State state = unicode.getField(&py_unicode_v::o_state);
+        //if (state.kind != 1 || state.compact != 1) {
+        //    throw InvalidRemoteObject();
+        //}

         len = unicode.getField(&py_unicode_v::o_length);
         buffer.resize(len);
```

That'll lead to problems if there's any non-ASCII strings, but we can circle back to that later. After applying that diff, I get a segfault (!!) in code that you changed in `pystack::getLocationInfo`. Adding this patch:
```diff
diff --git a/src/pystack/_pystack/pycode.cpp b/src/pystack/_pystack/pycode.cpp
index fc490e3..874fa25 100644
--- a/src/pystack/_pystack/pycode.cpp
+++ b/src/pystack/_pystack/pycode.cpp
@@ -133,6 +133,7 @@ getLocationInfo(
                 tlbc_entries + sizeof(tlbc_size),
                 tlbc_size * sizeof(uintptr_t),
                 vec.data());
+        LOG(DEBUG) << "tlbc_index=" << tlbc_index << " tlbc_size=" << tlbc_size;
         uintptr_t code_adaptive_actual = vec[tlbc_index];
         ptrdiff_t addrq =
                 (reinterpret_cast<uint16_t*>(last_instruction_index)
```
prints this out right before the crash:
```
DEBUG(process_remote): tlbc_index=-738076544 tlbc_size=0
```

Both of those look wrong, but it's not crashing on the first frame it processes, it's crashing on the 6th... Ah, it looks like we're trying to get the code object for a shim frame, which is just nonsense. This patch fixes that:
```diff
diff --git a/src/pystack/_pystack/pyframe.cpp b/src/pystack/_pystack/pyframe.cpp
index 3b81beb..6276952 100644
--- a/src/pystack/_pystack/pyframe.cpp
+++ b/src/pystack/_pystack/pyframe.cpp
@@ -29,10 +29,10 @@ FrameObject::FrameObject(
     if (d_is_shim) {
         LOG(DEBUG) << "Skipping over a shim frame inserted by the interpreter";
         next_frame_no = frame_no;
+    } else {
+        d_code = getCode(manager, frame);
     }

-    d_code = getCode(manager, frame);
-
     auto prev_addr = frame.getField(&py_frame_v::o_back);
     LOG(DEBUG) << std::hex << std::showbase << "Previous frame address: " << prev_addr;
     if (prev_addr) {
```

OK, so with those patches applied, things seem to actually work, but we will need to refactor things to allow for a different unicode state layout in free threaded vs non free threaded.

ZeroIntensity:
IIRC, dictionaries have a different layout in FT vs non-FT too.

godlygeek:
> IIRC, dictionaries have a different layout in FT vs non-FT too.

Looks like any differences there are covered by the debug offsets. After making the two fixes I identified yesterday (not trying to extract code objects from shim frames, and handling the state flags of `PyUnicode` having a different layout in 3.14t) the test suite passes.

We do probably need to add some more tests to prove that the thread-local bytecode stuff is working properly, though. I doubt that any of our existing tests force specialization in a non-main thread.

godlygeek:
The main thread's TLBC pointer points to the co_adaptive in the code object. If we're only testing in the main thread, there's no proof that we're not just using the bytecode for the main thread unconditionally.

pablogsal (approved):

PR #292, opened 2026-05-18
Fix security-related issues in github actions flagged by zizmor

### LINKED ISSUE #287: Fix problems reported by zizmor
Use [zizmor](https://zizmor.sh) to identify potential vulnerabilities in this project's CI setup, then fix them.

Once the report is clean, add zizmor to our CI with https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action

peytondmurray:
**Issue number of the reported bug or feature request:** Closes #287.

**Describe your changes**
This PR addresses a number of security-related issues in the github actions (and dependabot config) that were flagged by zizmor. Most of the changes are:

- Pinning action versions to commit hashes
- Disable persisting credentials in workflows after checkout
- The default permissions are now set as conservatively as possible

This PR also adds a zizmor action to ensure these issues don't arise in the future.

Note that the container images used during the wheel builds workflow remain unpinned. After weighing the dangers of running a compromised container here, it was decided that the effort of keeping pinned containers up to date was not worth the effort. As a result we have zizmor configured to ignore these rule violations here.

**Testing performed**
This will be tested by running the CI in this PR.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,5 +1,7 @@
 name: Wheels
 
+permissions: {}
+
 on:
   push:
   pull_request:
@@ -19,12 +21,14 @@ jobs:
     name: Build source distribution
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Build sdist
         run: pipx run build --sdist
 
-      - uses: actions/upload-artifact@v7
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: sdist
           path: dist/*.tar.gz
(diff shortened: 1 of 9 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -304,11 +321,13 @@ jobs:
       fail-fast: false
 
     container:
-      image: fedora
+      image: fedora@f717d3f59ea0
```
lkollar: Looks like this format isn't valid and the build fails.

We chatted about this with @godlygeek and it would be good to leverage Dependabot for keeping the Docker images up to date as well. It seems it's [supported](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker).

peytondmurray: Great, I've added the `Docker` section to the dependabot config. It sounds like we also need labels [on each image](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker) if we want downstream projects that rely on dependabot to pull in metadata about new releases. I've added

```dockerfile
LABEL org.opencontainers.image.source="https://github.com/bloomberg/pystack"
```

to the last image to do this.


The other thing is that the Dependabot config only works if

> maintainers must tag the repository with the same tags as the published Docker images. For an example, see the [dependabot-fixtures/docker-with-source](https://github.com/dependabot-fixtures/docker-with-source) repository.

I assume they're talking about git tags, right?

sarahmonod:
I know this was already the case before this PR, but I don't like the fact that the upload and download actions are of different versions.

Can you please check if the latest version for both of them works together?

peytondmurray:
Sure! I figured I'd let dependabot take care of it, but I'd be happy to update _all_ of the actions, which probably makes the most sense.

peytondmurray:
Okay, after a few minutes investigation it seems like there is currently a version discrepancy between actions/upload-artifact and actions/download-artifact. So this is fine as is, and in fact it is configured in accordance with the download-artifact usage instructions: https://github.com/actions/download-artifact#download-multiple-filtered-artifacts-to-the-same-directory

I did take the opportunity to bump the setup-uv action to v8, though.

godlygeek (approved):
Thanks for these cleanups, @peytondmurray!

PR #294, opened 2026-05-18
Pin the pre-commit hook hashes

### LINKED ISSUE #293: Use prek instead of pre-commit
Currently, `.pre-commit-config.yaml` uses git tags rather than hashes for a number of hooks. These are vulnerable to the same type of "mutable git ref" attacks that unpinned github actions are. We've already seen attempts to exploit this in the wild on the Jupyter project.

Let's update these with a `pre-commit autoupdate --freeze` to pin these.

peytondmurray:
**Issue number of the reported bug or feature request:** Closes #293.

**Describe your changes**
This PR pins (and updates) the pre-commit hooks using commit hashes rather than (potentially mutable) git refs. For future reference, the command used here was `pre-commit autoupdate --freeze`.

**Testing performed**
No testing required.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c  # frozen: v6.0.0
     hooks:
       - id: check-added-large-files
       - id: check-json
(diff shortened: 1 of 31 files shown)


peytondmurray:
It looks like the newly updated pre-commit hooks resulted in a few formatting changes, some of which hit example python files used by tests. This means that there are some line number changes for a few of the integration tests. These should now be up to date with the latest changes :+1:

peytondmurray:
Does it make sense to write a news entry for this?

sarahmonod:
> Does it make sense to write a news entry for this?

Nope, I've added the skip news label for it.

sarahmonod (approved):

PR #299, opened 2026-05-18
Replace `pre-commit` with `prek`

### LINKED ISSUE #293: Use prek instead of pre-commit

sarahmonod:
Relates-to: #293


godlygeek (approved):
LGTM

PR #308, opened 2026-05-19
Upgrade CMAKE_CXX_STANDARD to 20

scopreon:
*Issue number of the reported bug or feature request: NAN*

**Describe your changes**
Upgrade supported C++ version from 17 to 20

**Testing performed**
CI should pass

**Additional context**


sarahmonod:
This looks good to me in principle, but I'm not sure why we were using C++17 to begin with, so I'd like Matt's input

godlygeek:
I'm reasonably convinced that our CI would catch any problems with this. The fact that we were able to successfully produce `manylinux_2_17` and `musllinux_1_2` wheels means that this doesn't introduce any runtime dependencies on newer versions of libstdc++ or libc than we're able to depend upon, which would have been my only concern about upgrading.

So, I think this is safe... any objections, @pablogsal?

pablogsal:
No concerns, we have done this with cpp17 in the past iirc with no problems as well

sarahmonod (approved):
LGTM, thanks @scopreon !

PR #309, opened 2026-05-20
Exit with a clear message when ptrace_scope blocks the test suite

### LINKED ISSUE #307: Produce a helpful error message if `ptrace_scope` is 0

If a user tries to run the integration tests, they can hang indefinitely if

```bash
cat /proc/sys/kernel/yama/ptrace_scope
```

prints `1`.

I'd love if pystack generated an error message with a solution instead of hanging indefinitely. You can fix this by doing

```bash
sudo bash -c 'echo 0 >> /proc/sys/kernel/yama/ptrace_scope'
```

It would be nice to do this check before running tests in conftest.py so that folks know they need to deal with this before spending time on it.

_No response_

c-tonneslan:
Closes #307.

When Yama's `ptrace_scope` is set to anything other than `0`, pystack can't attach to a running process, so the integration tests hang instead of failing with a useful message.

`tests/conftest.py` already does a `pytest_sessionstart` preflight for `gcore`, so I added a `ptrace_scope` check right next to it: read `/proc/sys/kernel/yama/ptrace_scope`, and if it's non-zero, `pytest.exit` with the command to fix it (`sudo sysctl kernel.yama.ptrace_scope=0`).

The read is wrapped in `try/except OSError`, so on a system where that file doesn't exist (non-Linux, or a kernel built without Yama) the check is just a no-op.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,6 @@
+import os
 import shutil
+from pathlib import Path
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


godlygeek:
Ah - we need one more check here. We can bypass this if the user is root, or if we have `CAP_SYS_PTRACE`.

It seems a bit annoying to check for the latter, so let's just check for root for now, and we'll cross the bridge of non-root users with debugging privileges needing to run the pytest test suite when we come to it...

godlygeek (approved):
Looks good to me after adding a special case for root - thanks @c-tonneslan!

PR #315, opened 2026-05-29
Update elfutils version and use ubuntu apt packages in dockerfile

### LINKED ISSUE #314: Update to the latest elfutils
Update the cibuildwheel configuration in pyproject.toml to use elfutils 0.195 instead of 0.193.

Stop building elfutils from source in Dockerfile and just use Ubuntu's libelf-dev and libdw-dev instead. The latest Ubuntu version has a newer elfutils than the one we've been building from source.

mbmtbtbc:
Issue number of the reported bug or feature request: fixes #314 

**Describe your changes**
Replace the custom `elfutils` build stage in the Dockerfile with Ubuntu's packaged `libdw-dev` and `libelf-dev` dependencies.

Removed the `elfutils_builder` stage from the Dockerfile
Removed the download and compilation of elfutils 0.193
Added `libdw-dev` and `libelf-dev` to the Ubuntu package list
Removed the copy step that imported the custom-built libraries
Updated documentation referencing the source build of elfutils

**Testing performed**
Successfully built the Docker image
Verified `pkg-config` detects both `libdw` and `libelf`
Successfully built and installed pystack inside the container using the distro-provided libraries

**Additional context**
`Ubuntu 26.04` currently provides `elfutils/libdw` version `0.194` rather than `0.195`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Once the codespace is ready, all dependencies will be installed and hooks
 will be configured automatically.
 
 Note that the dev container is built from the project's `Dockerfile`, which
-compiles `elfutils` from source, so the first build takes a few minutes.
+uses the distribution-provided `libdw` and `libelf` development packages.
 
 pystack requires elevated ptrace permissions to inspect processes — the dev
 container already handles this via `--cap-add=SYS_PTRACE` and
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,4 @@
-# Stage 1: Elfutils build stage
-FROM ubuntu:26.04 AS elfutils_builder
-ARG DEBIAN_FRONTEND=noninteractive
-ENV VERS=0.193
-
-# Install elfutils build dependencies
-RUN apt-get update \
-    && apt-get install -y --force-yes --no-install-recommends software-properties-common gpg-agent \
-        build-essential \
-        libzstd-dev \
-        ca-certificates \
-        curl \
-        lsb-release \
-        bzip2 \
-        zlib1g-dev \
-        zlib1g-dev:native \
-        libbz2-dev \
-        liblzma-dev \
-        gettext \
-        po-debconf \
-        gawk \
-        libc6-dbg \
-        flex \
-        bison \
-        pkg-config \
-        libarchive-dev \
-        libcurl4-gnutls-dev \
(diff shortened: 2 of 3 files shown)


mbmtbtbc:
I worked on the Dockerfile changes and replaced the custom elfutils build with Ubuntu's `libdw-dev` and `libelf-dev` packages.

One thing I noticed while validating the change is that Ubuntu 26.04 currently provides elfutils/libdw version `0.194` rather than `0.195`:

```bash
pkg-config --modversion libdw
0.194

pkg-config --modversion libelf
0.194
```

The Docker image builds successfully and `pystack` also builds and installs successfully against the distro-provided libraries.

Is `0.194` acceptable for the Docker image, or would you prefer a different approach to obtain `0.195`?

godlygeek:
> Is `0.194` acceptable for the Docker image, or would you prefer a different approach to obtain `0.195`?

It's fine for us to use 0.194 in the Docker image. Since the Docker image is used only for testing, the only reason we'd need to use a newer version there is if we want to use a feature that's only available in a later version, or if we're affected by a bug in 0.194 that's fixed in a later version. If either of those ever happens, we can reintroduce the build from source at that point.

This PR will need to be rebased, though - there's a minor merge conflict between it and #313. Also, it's missing the Signed-off-by line in the commit message body - see CONTRIBUTING.md

More worryingly, it looks like the build of elfutils 0.195 is failing in the manylinux container that cibuildwheel is using...

mbmtbtbc:
Understood! Here's what I'll do:

1. Rebase onto main to resolve the `Dockerfile` conflict
2. Amend the commit to add the `Signed-off-by` line
3. Dig into the manylinux build log to figure out why 0.195 is failing there

Will push an update once I've sorted it out.

godlygeek:
Sorry for taking so long to get around to investigating this.

> 3\. Dig into the manylinux build log to figure out why 0.195 is failing there

The issue turned out to be that the manylinux2014 container has a version of `make` that isn't new enough (make 3.x instead of 4.x). Installing a newer `make` fixes the issue.

Other than that, your changes all look good to me. I've rebased this, split it into two commits, and updated the commit messages.

Thanks for the contribution, @mbmtbtbc!

godlygeek (approved):

mbmtbtbc:
That's awesome, thanks! It was a meaningful learning experience!

PR #329, opened 2026-07-10
Resolve native symbols for containerized python applications

### LINKED ISSUE #327: Resolve native symbols for containerized Python processes using the target process root (via /proc/<pid>/root)

  Yes.

  When `pystack remote --native <pid>` attaches to a Python process running inside a container runtime such as
  Enroot, Python frames can still be visible, but native C/C++ symbol resolution fail for shared
  libraries whose mapped paths only exist inside the container filesystem. 

  A simplified example:

  ```console
  Python process:
    loads libnative_hello.so with ctypes

  Inside container:
    /tmp/pystack-native-symbols-current/libnative_hello.so  => exists

  Host running PyStack:
    /tmp/pystack-native-symbols-current/libnative_hello.so  => does not exist
```

The host can still see the target process mappings:

```console
  grep libnative_hello.so /proc/<pid>/maps

  Example mapping:

  ... /tmp/pystack-native-symbols-current/libnative_hello.so
```

But resolving that path directly from the host fails:

```console
  test -r /tmp/pystack-native-symbols-current/libnative_hello.so => not readable from host
```

The same file is reachable through the target process root:

```console
  test -r /proc/<pid>/root/tmp/pystack-native-symbols-current/libnative_hello.so => readable
```

When PyStack resolves native symbols for a live process, if opening a mapped ELF path from `/proc/<pid>/maps` fails, retry the lookup through the target process root `/proc/<pid>/root/<mapped-path>`.

For example, if `/proc/<pid>/maps` contains:

```console
  /tmp/pystack-native-symbols-current/libnative_hello.so
```

  and the host cannot open that path directly, PyStack should try:

```console
  /proc/<pid>/root/tmp/pystack-native-symbols-current/libnative_hello.so
```

This would allow PyStack to resolve native symbols for libraries that are visible to the target process but not directly visible in the host filesystem namespace.

The fallback should only be used when direct path lookup fails, so normal host behavior remains unchanged.

 1. Require users to bind-mount container paths at the same location on the host.

  This works in some controlled cases, but not feasible with the large container stack used today.

2. Ask users to run PyStack inside the same container.

  This can work for simple cases, but it is not always compatible with attach workflows where the debugger/profiler
  runs on the host or from a scheduler/tooling process outside the container

pramodk:
**Describe your changes**

This change fixes native symbol lookup for Python processes running inside container runtimes such as [Enroot](https://github.com/nvidia/enroot).

When the libdwfl lookup cannot open an absolute module path from `/proc/<pid>/maps`, this PR now retries that same path through the target process root:

```text
  /proc/<pid>/root/<mapped-path>
```

The fallback is only used after build-id lookup and normal host path lookup fail, so existing host behavior remains unchanged.

Fixes #327.
  
**Testing performed**

  - Built a PyStack wheel from this PR.
  - Validated with a standalone ctypes repro:
      - Host process: upstream and patched PyStack resolve Python and native frames.
      - Enroot process with current main: Python frames are visible, native symbol is missing.
      - Enroot process with this PR: Python frames are visible, native symbol is resolved.
      
**Additional context**

The issue is specific to native C/C++ symbol lookup. Python frames can still be visible because PyStack can inspect the target process, but native library paths from `/proc/<pid>/maps` may exist only inside the container filesystem. As containers are used extensively in certain domains like AI/ML/Data Science, it would be really helpful to fix this.

### DIFF
--- a/src/pystack/_pystack/elf_common.cpp
+++ b/src/pystack/_pystack/elf_common.cpp
@@ -5,6 +5,7 @@
 #include <iomanip>
 #include <iostream>
 #include <string>
+#include <sys/stat.h>
 #include <utility>
 
 #include "compat.h"
@@ -13,6 +14,58 @@
 namespace pystack {
 
 using file_unique_ptr = std::unique_ptr<FILE, std::function<int(FILE*)>>;
+
+namespace {
+
+int
+set_process_pid(
+        Dwfl_Module* mod __attribute__((unused)),
+        void** userdata,
+        const char* name __attribute__((unused)),
+        Dwarf_Addr start __attribute__((unused)),
+        void* arg)
+{
+    *userdata = arg;
+    return DWARF_CB_OK;
+}
+
+int
+find_elf_through_process_root(void** userdata, const char* modname, char** file_name)
+{
+    if (userdata == nullptr || *userdata == nullptr || modname == nullptr || modname[0] != '/') {
+        return -1;
+    }
+
+    const auto pid = *static_cast<const pid_t*>(*userdata);
+    if (pid <= 0) {
+        return -1;
+    }
+
+    const std::string rooted_path =
+            "/proc/" + std::to_string(pid) + "/root" + std::string(modname);
+    int fd = open(rooted_path.c_str(), O_RDONLY);
+    if (fd < 0) {
+        return -1;
+    }
+
+    struct stat file_stat;
+    if (fstat(fd, &file_stat) != 0 || !S_ISREG(file_stat.st_mode)) {
+        close(fd);
+        return -1;
+    }
+
+    if (file_name != nullptr) {
+        *file_name = strdup(rooted_path.c_str());
+        if (*file_name == nullptr) {
+            close(fd);
+            return -1;
+        }
+    }
+
+    return fd;
+}
+
+}  // namespace
 
 int
 pystack_find_elf(
@@ -31,10 +84,14 @@ pystack_find_elf(
         return ret;
     }
     ret = dwfl_linux_proc_find_elf(mod, userdata, modname, base, file_name, elfp);
-    if (file_name == nullptr) {
+    if (ret < 0) {
+        ret = find_elf_through_process_root(userdata, modname, file_name);
+    }
+    if (ret < 0) {
         LOG(DEBUG) << "Could not locate debug info for " << the_modname;
     } else {
-        LOG(DEBUG) << "Located debug info for " << the_modname << " by path in " << *file_name;
+        const char* the_filename = (file_name == nullptr || *file_name == nullptr) ? "???" : *file_name;
+        LOG(DEBUG) << "Located debug info for " << the_modname << " by path in " << the_filename;
     }
     return ret;
 }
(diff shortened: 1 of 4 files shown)


src/pystack/_pystack/elf_common.cpp
```diff
@@ -31,10 +70,14 @@ pystack_find_elf(
         return ret;
     }
     ret = dwfl_linux_proc_find_elf(mod, userdata, modname, base, file_name, elfp);
-    if (file_name == nullptr) {
+    if (ret < 0) {
+        ret = find_elf_through_proc_pid_root(userdata, modname, file_name);
+    }
```
godlygeek, line 108: This only falls back to checking through `/proc/$pid/root` if the lookup on the host filesystem failed, but that seems wrong to me - isn't it possible that we've found entirely the wrong file if we looked on the host filesystem first? I would think that we'd want to always go through `/proc/$pid/root` whenever the mount namespace doesn't match. Or always, really? I can't think of any reason we wouldn't want to always resolve relative to it...

godlygeek:
This will definitely need an automated integration test added to the test suite, also!

src/pystack/_pystack/elf_common.cpp
godlygeek: In fact, I wonder whether this is something we should be implementing in pystack at all. Perhaps libdwfl should automatically be using /proc/$pid/root whenever it resolves files from /proc/$pid/map...

pramodk: @godlygeek - Sorry for delay in response during last month. 

I have pushed new commits to address some of these points:

- Use `/proc/$pid/root` for module lookup. I have kept the old path as a fallback in case there are corner cases where this is required.
- Integration test that exercises this behavior with separate namespaces.

Regarding `libdwfl` taking care of this, I have checked the libdwfl implementation and asked on the developer list about possible fixes to support this scenario. I will see if there is any response or concrete proposal.

Until that type of support is generally available, it would be helpful to have this fix in PyStack?

I have tested this in a [fork here](https://github.com/pramodk/pystack/pull/5). So hopefully the tests will pass after CI is approved to run.

pramodk: > Regarding libdwfl taking care of this, I have checked the libdwfl implementation and asked on the developer list about possible fixes to support this scenario. I will see if there is any response or concrete proposal.

@godlygeek : I didn't receive any feedback from libdwfl. I will poke again but could you check this updated PR and see if how the changes+test organized looks good or any changes would be needed?

godlygeek: Looks reasonable at a glance. I'll take a closer look sometime soon. I don't like `set_module_process_pid_userdata` - it feels like a hack to need to iterate over each of the modules and individually set the PID as user-data on each of them. But I don't see any better option in the libdwfl API, so I guess our options are either that or a global variable. Yuck. I guess it's fine as it is.

The tests look more complex than I was hoping for, but I haven't had a chance to review in detail yet.

I do expect I'll land this change, possibly with some massaging - and there's some things I'll need to read more carefully to make sure I understand them.

godlygeek:
I'm still a bit reluctant to merge a change in pystack for this. I'm pretty convinced that the right place to solve it is elfutils, so that it benefits every user of libdw (or at least, every user attaching to remote processes).

I've filed https://sourceware.org/bugzilla/show_bug.cgi?id=34643 and I plan to wait a bit to see if the elfutils maintainers agree or disagree that this is something libdw ought to handle itself.

pramodk:
Initially I posted this on the elfutils-devel mailing [list here](https://sourceware.org/pipermail/elfutils-devel/2026q3/009573.html) and didn't get response. Should have created bug report instead. 

Sounds good, lets wait and see what the elfutils devs says.

PR #331, opened 2026-07-12
ci: Tag stable abi wheels as abi3

godlygeek:
We've been limiting ourselves to the stable ABI for Python 3.12+, but we haven't been tagging those stable ABI wheels as `abi3`.

Fix this, allowing the wheels built for Python 3.12 to be used for 3.13, 3.14, and 3.15 as well.

Note that nanobind does not yet support `abi3t`, so free-threaded wheels are still built per-version.


godlygeek:
This speeds CI up by about 10%, by skipping 3 builds of the extension module for each arch/libc.

sarahmonod (approved):

PR #352, opened 2026-09-01
Avoid repeated cache fills for unreadable memory maps

pablogsal:
ProcessMemoryManager retries whole-map cache fills on every access when a mapping contains an unreadable page. Remember mappings whose cache fill fails and read requested ranges directly on subsequent accesses, avoiding repeated failed reads and temporary allocations.


godlygeek:
@pablogsal Mind adding a Signed-off-by line, please?

godlygeek (approved):

PR #359, opened 2026-09-17
Drop fallback to `--version` for version detection

godlygeek:
This fallback presumed that the executable is a Python interpreter that takes the Python interpreter's typical arguments, but could fail catastrophically if the interpreter were to be some executable that embeds a Python interpreter and that does something destructive or unexpected when run with `--version` (likely because it ignores its command line arguments entirely).


blin (approved):
LGTM.

If we want to claw back some detection power for corrupted cores of pre 3.13 embedded statically linked pythons we can add a .rodata scan (I think this is the only case where we would actually gain anything)? But even if that is feasible/useful it is beyond the scope of this PR.

